PR #6, opened 2024-10-17
fix: Unexpected sensor states

heinoskov:
added few other sensors states and boolean value for them


woopstar:
@heinoskov  I merged this, but just saw it contained a ready. Should that be True and not False?

heinoskov:
For charging status it should be false as its only ready to charge but not charging

PR #19, opened 2024-11-09
fix: add charge at best time for battery

woopstar:
This will add hsem_huawei_solar_batteries_grid_charge_cutoff_soc for config flow and also add a new function to calculate the best time to charge the battery.


woopstar:
This PR will start charging the battery during winter time.

```python
# Charge the battery when it's winter/spring and prices are high
if now.month in DEFAULT_HSEM_MONTHS_WINTER_SPRING:
    # find best time to charge the battery at night
    if now.hour >= 0 and now.hour < 6:
        await self.async_find_best_time_to_charge(0,6)

    # find best time to charge the battery at day
    if now.hour >= 12 and now.hour < 17:
        await self.async_find_best_time_to_charge(12,17)
```

it will charge the battery at the cheapest time between 00 and 06, and also between 12 and 17.

PR #425, opened 2026-05-16
fix(planner): overhaul MILP optimizer, add partial-SoC, discharge concentration, time-discount, fix 11 bugs (A-K) + 3 review issues

woopstar:
# Summary

Comprehensive overhaul of the HSEM planner and MILP optimizer across 21 commits. Covers cycle cost correction, terminal-SoC accounting, mutual exclusion, candidate diversity, logging improvements, and 11 code-review bugs (A--K).

### Cycle cost (cost_function.py, utils/misc.py)
- `2x` factor added to `_resolve_cycle_cost` denominator: `purchase_price / (2 * usable_kwh * expected_cycles)`
- `calculate_recommended_threshold` updated to use same `2x` formula + new `capacity_loss_pct` parameter (default 30%)
- Removed deprecated `conversion_loss` param from `SensorConfig` and `calculate_recommended_threshold`

### MILP optimizer (milp_optimizer.py)
- **Bug 9**: Negative `net_load` infeasibility -- added explicit `pv[t]` LP variable bounded to solar surplus
- **Bug 10**: Solver timeout -- increased from 0.5s to 2.0s for 192-slot horizons
- **Bug 11**: Magnitude-based tiebreak -- replaced with price comparison
- **Bug A**: Cycle cost (rejected, original correct -- `cycle_cost_per_kwh` includes `2x` denominator)
- **Bug B**: `hold[t]` variable (6*m) replaced with terminal-SoC opportunity cost on `ec[t]/ed[t]` coefficients (5*m). Mirrors `terminal_soc_value` in `cost_function.py`
- **Bug C**: Mutual exclusion at negative prices -- added `ec[t]/max_charge + ed[t]/max_dis <= 1` inequality to `A_ub`
- **Bug J**: Efficiency-aware mutex resolution -- checks round-trip profitability before picking charge/discharge
- **Bug K**: `hold_reward` NaN edge case (already fixed by Bug B removal)

### Candidates (candidate_generator.py, candidate_selector.py)
- **Bug D**: Aggressive charge/discharge ordering -- charge first, discharge second, post-hoc cleanup removes charge after first discharge
- **Bug E**: Deduplicate soc_plan candidates -- skip near-identical `charge_target` values (0.05 kWh tolerance)
- **Bug F**: Solar sort key -- sort by `estimated_net_consumption` asc (most PV surplus first) instead of import price
- **Bug G**: Charge efficiency -- `charge_efficiency_pct` param; `charge_target /= charge_eff`
- **Bug H**: `_cost` as proper dataclass field on `CandidatePlan`; removed all `# type: ignore[attr-defined]`
- **Bug I**: `ForceBatteriesDischarge` now included in `replacement_price_from_next_discharge` via `_DISCHARGE_RECS`
- Baseline positional tie-break removed; uses `total_cost + name` for deterministic ordering
- `top_n` derived dynamically: `ceil(usable_kwh / max_discharge_per_slot)`
- `GAP_THRESHOLD` derived dynamically from `interval_minutes + 5`
- Aggressive discharge slot count derived dynamically (not hardcoded 3)
- Charge-schedule capacity cap: total charge capped to battery remaining capacity across discharge-window occurrences

### Partial-SoC (BatPred) candidates
- 6 charge-fraction candidates: 25%, 50%, 75%, 100%, 125%, fill-to-max
- Each charges a different fraction of the energy needed for upcoming discharge windows
- Discharge-fraction mode activates when battery is already well-charged

### Discharge concentration
- `concentrate_discharge_on_expensive_slots` clears cheap discharge slots the battery cannot fully serve
- SoC simulation respects `BatteriesWaitMode` -- preserves stored energy for expensive future slots

### Time-discount factor
- Per-hour exponential discount (default 0.995) on the selector score (not `total_cost`)
- Applied to MILP objective as well, using per-slot midpoint discount
- Terminal-SoC credit also discounted at horizon-end rate

### Dedicated HSEM log file
- `RotatingFileHandler` writes to `hsem.log` (10 MB x 5 files) in HA config dir
- `HSEM_LOGGER.propagate = False` keeps planner debug out of `home-assistant.log`
- Init/teardown offloaded to `ThreadPoolExecutor` to avoid blocking HA event loop

## Breaking Changes

- Removed `hsem_batteries_conversion_loss` config option (replaced by separate charge/discharge efficiencies)
- Removed `hsem_batteries_schedule_*_min_price_difference` config options (auto-calculated from depreciation)
- `batteries_conversion_loss` attribute removed from sensor state
- `calculate_recommended_threshold` signature changed (removed `conversion_loss`, added `capacity_loss_pct`)

## Verification

- 1839 tests pass (3 xfailed)
- Lint and quality pass
- `n_vars = 5*m` in MILP with correct offsets
- No `# type: ignore[attr-defined]` comments remain
- Solver time stays under 2.0s with `5*m` variables

Fixes #9, #10, #11


woopstar (commented):
## Code Review — PR #425

God PR. Bug 5 (`_AGGRESSIVE_DISCHARGE_SLOTS`) og `top_n` er nu korrekt dynamiske — det er de vigtigste rettelser. Et problem er stadig urettet.

---

### Urettet: Tie-break favoriserer baseline uanset hvad

I `candidate_selector.py` linje ~148:

```python
# Sort by selector score ascending; baseline wins ties (it comes first)
eligible_sorted = sorted(
    eligible,
    key=lambda c: (
        c._cost.score,
        # Stable tie-break: baseline index is 0, so it wins ties
        next(
            (i for i, x in enumerate(candidates) if x is c),
            len(candidates),
        ),
    ),
)
```

Kommentaren siger det selv: *"baseline wins ties"*. Men `candidates`-listen er genereret med baseline som index 0, og de 6 nye partial-SoC kandidater (25%, 50%, 75%, 100%, 125%, fill) ligger på index 2–7. Det betyder en `partial_50pct`-kandidat der scorer identisk med baseline **aldrig kan vinde** — ikke fordi den er dårligere, men fordi dens list-index er højere.

Med 6 nye kandidater i denne PR er problemet nu større end før. En `partial_100pct`-kandidat der koster præcis det samme som baseline men lader batteriet mere op til næste dag, vil konsekvent blive afvist på position alene.

**Foreslået rettelse:**

```python
# Sort by score ascending, then total_cost ascending (real savings),
# then name for determinism. Baseline gets no positional advantage.
eligible_sorted = sorted(
    eligible,
    key=lambda c: (
        c._cost.score,       # primær: laveste selector score
        c._cost.total_cost,  # sekundær: laveste reelle totalomkostning
        c.name,              # tertiær: deterministisk, ingen positional bias
    ),
)
```

Dette er en one-liner rettelse. Den fjerner baseline-favoriseringen uden at introducere ikke-determinisme.

---

### Observation: rejected-plan reason er misvisende ved uafgjort

Når en kandidat taber på tie-break, skriver koden:

```python
reason = (
    f"Tied or marginally worse score "
    f"({candidate_score:.4f}); baseline preferred."
)
```

Efter rettelsen ovenfor er teksten `"baseline preferred"` ikke længere korrekt — det er nu `total_cost` der afgør. Opdater reason-teksten til:

```python
reason = (
    f"Tied score ({candidate_score:.4f}); "
    f"higher total_cost than winner ({winner._cost.total_cost:.4f})."
)
```

---

### Alt andet ser godt ud ✅

- Dynamisk `top_n` via `ceil(usable_kwh / max_discharge_per_slot)` — korrekt
- Dynamisk `GAP_THRESHOLD` via `interval_minutes + 5` — korrekt
- Dynamisk aggressive discharge slots — korrekt og fjernet dead constant
- MILP PV-variabel og price-based tiebreak — solid fix
- 1836 tests passer — ingen regressioner

Merge gerne efter tie-break-rettelsen.

woopstar:
## Feature request: Time-discount factor on cost function score

### Problem

`score_plan()` currently weights every slot equally regardless of when it occurs. A saving 47 hours from now counts the same as a saving 1 hour from now. This is wrong for two reasons:

1. **Forecast uncertainty grows with time.** Spot prices, solar yield, and consumption 40+ hours out are significantly less reliable than the next few hours.
2. **Optimistic distant plans beat conservative near-term plans unfairly.** The selector can pick a candidate that looks cheap because it relies on a perfectly-shaped discharge window 2 days from now, even though the baseline plan is more robust and almost as cheap.

### Proposed fix — exponential discount on `score` only

Add a per-slot discount multiplier to the `score` objective (not to `total_cost` — that must remain auditable real money):

```
discount(t) = discount_rate ^ hours_from_now(t)
```

where `discount_rate` is a value slightly below 1.0. A recommended default is **0.995 per hour**, which gives:

| Horizon | Factor |
|---|---|
| 1 hour | 0.995 |
| 6 hours | 0.970 |
| 24 hours | 0.887 |
| 48 hours | 0.787 |

This means a saving 48 hours from now is only worth ~79% of a saving right now in the selector's eyes — still significant, but no longer equal.

### Implementation

**1. Add `time_discount_rate` to `CostWeights`**

```python
@dataclass
class CostWeights:
    # ... existing fields ...
    time_discount_rate: float = 0.995  # per-hour discount on score terms
```

A value of `1.0` disables the discount entirely (backward-compatible default).

**2. Compute `hours_from_now` per slot in `score_plan()`**

```python
# At the top of the slot loop — requires `now` to be provided.
# Falls back to no discount when `now` is None.
if now is not None and weights.time_discount_rate < 1.0 - 1e-9:
    slot_mid = slot.start + (slot.end - slot.start) / 2
    hours_ahead = max((slot_mid - now).total_seconds() / 3600.0, 0.0)
    discount = weights.time_discount_rate ** hours_ahead
else:
    discount = 1.0
```

**3. Apply discount to the money terms inside the slot loop**

```python
# 1. Import cost
if slot.grid_import_kwh > 1e-9:
    import_cost += slot.grid_import_kwh * imp_price * discount   # <- discount added

# 2. Export revenue
if slot.grid_export_kwh > 1e-9:
    export_revenue += slot.grid_export_kwh * exp_price * discount  # <- discount added

# 3. Conversion loss cost
if slot.batteries_charged > 1e-9 and charge_loss_fraction > 1e-9:
    lost_kwh_charge = slot.batteries_charged * charge_loss_fraction
    conversion_loss_cost += lost_kwh_charge * imp_price * discount   # <- discount added
if slot.batteries_discharged > 1e-9 and discharge_loss_fraction > 1e-9:
    lost_kwh_discharge = slot.batteries_discharged * discharge_loss_fraction
    conversion_loss_cost += lost_kwh_discharge * imp_price * discount  # <- discount added

# 4. Cycle depreciation
if throughput_kwh > 1e-9 and cycle_cost_kwh > 1e-9:
    cycle_cost_total += throughput_kwh * cycle_cost_kwh * discount   # <- discount added

# 5-7. SoC penalty, grid limit, override — also discounted (optional)
#      These are selector-only penalties so discounting them is fine.
soc_penalty += weights.soc_low_penalty_weight * violation**2 * discount
```

**Important:** `total_cost` must NOT be discounted. It is computed from the raw (non-discounted) `import_cost`, `export_revenue`, etc. Only `score` uses the discounted values. This keeps `total_cost` auditable as real money while the selector becomes time-aware.

The cleanest way to implement this is to maintain two running totals per term — one raw (for `total_cost`) and one discounted (for `score`).

**4. Also apply to `milp_optimizer.py`**

The MILP objective vector `c_obj` should also include the discount factor so the LP solution and the cost function stay consistent:

```python
# In solve_milp(), after computing hours_ahead per slot:
discount = discount_rate ** hours_ahead[t]   # per-slot scalar

c_obj[ec_off + t] = (cycle_cost_per_kwh + charge_loss * p_imp[t]) * discount
c_obj[ed_off + t] = (cycle_cost_per_kwh + discharge_loss * p_imp[t]) * discount
c_obj[gi_off + t] = p_imp[t] * discount
c_obj[ge_off + t] = -p_exp[t] * discount
```

`discount_rate` should be passed through from `PlannerInput` → `generate_candidates()` → `solve_milp()`.

### Files to change

| File | Change |
|---|---|
| `planner/cost_function.py` | Add `time_discount_rate` to `CostWeights`, apply per-slot discount in `score_plan()` |
| `planner/milp_optimizer.py` | Accept `discount_rate` param, apply to `c_obj` per slot |
| `planner/candidate_generator.py` | Pass `discount_rate` through to `solve_milp()` |
| `models/planner_inputs.py` | Add `time_discount_rate: float = 0.995` to `PlannerInput` |
| `planner/engine.py` | Pass `inp.time_discount_rate` into `CostWeights` and `generate_candidates()` |
| `tests/planner/test_cost_function.py` | Add test: discounted score < undiscounted score for distant-only plan |

### Acceptance criteria

- [ ] `CostWeights(time_discount_rate=1.0)` produces identical results to current behaviour (no regression)
- [ ] A plan with all savings in slot 0 scores better than an identical plan with all savings in slot 47
- [ ] `total_cost` is not affected by `time_discount_rate` (auditable money is undiscounted)
- [ ] MILP `c_obj` uses same discount vector as `score_plan()`
- [ ] Default `time_discount_rate=0.995` is documented in `CostWeights` docstring with the table above

woopstar:
## Feature request: Explicit `hold[t]` variable with hold-reward in MILP

### Problem

Slots that are neither charging nor discharging currently receive `recommendation = None`. The LP has no explicit concept of "hold the battery and wait for a better price". The hold decision emerges passively — `ec[t] = 0` and `ed[t] = 0` — purely because charge and discharge have higher costs. There is no positive signal rewarding the LP for preserving stored energy.

This causes two subtle issues:

1. **The LP is indifferent between hold and micro-discharge.** When `p_imp[t]` is moderate, `ed[t]` might get a small positive value (numerical noise or a genuine but barely-profitable discharge) when the correct answer is to hold and wait for a higher-price slot later in the horizon.
2. **SoC conservation has no objective weight.** The SoC bounds prevent violating capacity constraints, but within those bounds the LP has no preference for keeping energy in the battery over leaking it slowly across many low-value discharge slots.

### Proposed fix — explicit `hold[t]` variable

Add a `hold[t]` variable representing battery energy that is explicitly "reserved" this slot — not charged, not discharged, just held. Assign it a small **negative** objective coefficient (a reward) so the LP actively prefers holding over marginal actions.

**Variable layout after the change:**

```
x = [ec(0..m-1), ed(0..m-1), gi(0..m-1), ge(0..m-1), pv(0..m-1), hold(0..m-1)]
Offsets: ec_off=0, ed_off=m, gi_off=2m, ge_off=3m, pv_off=4m, hold_off=5m
n_vars = 6 * m
```

**Objective coefficient for `hold[t]`:**

```python
# Small reward for holding stored energy — discourages marginal discharge.
# Value should be small enough not to override a genuinely profitable charge
# or discharge, but large enough to break ties in favour of holding.
# Recommended: 0.001 * avg_import_price  (roughly 0.1% of import cost)
hold_reward = 0.001 * float(np.mean(p_imp[p_imp > 1e-9])) if np.any(p_imp > 1e-9) else 1e-4
c_obj[hold_off : hold_off + m] = -hold_reward  # negative = reward
```

**`hold[t]` bounds:**

```python
# hold[t] is bounded by current SoC estimate at slot t.
# Since SoC is a derived (not explicit) LP variable, use usable_kwh as
# a conservative upper bound. The SoC inequality constraints already
# prevent over-commitment.
+ [(0.0, usable_kwh)] * m   # hold[t]
```

**`hold[t]` does NOT enter the energy balance equation** — it is a purely objective-side variable that rewards inaction. It has no coupling to the SoC recurrence or the grid balance. This keeps the LP structure unchanged.

```python
# Energy balance: hold[t] has zero coefficient in A_eq
# No change to equality constraints needed
```

**`hold[t]` is coupled to the SoC via an inequality** — you can only hold what you have:

```python
# hold[t] <= soc[t-1]   (can't hold more than what's in the battery)
# Expressed as: hold[t] - cumsum(ec - ed)[0..t-1] <= current_kwh
# Add one row per slot to A_ub:
for t in range(m):
    row = np.zeros(n_vars)
    row[hold_off + t] = 1.0
    for k in range(t):           # soc[t-1] = current_kwh + sum(ec-ed)[0..t-1]
        row[ec_off + k] = -1.0
        row[ed_off + k] = 1.0
    A_ub_rows.append(row)
    b_ub_rows.append(current_kwh)
```

### Decode step — no change needed

`hold[t]` is an objective-only variable. The existing decode block reads `ec_sol` and `ed_sol` and writes recommendations — `hold_sol` does not need to be decoded into a recommendation. Slots where `ec[t] ≈ 0` and `ed[t] ≈ 0` continue to get `recommendation = None`, which maps to `BatteriesWaitMode` in the fill pass.

Optionally, log the total hold reward for debugging:

```python
hold_sol = result.x[hold_off : hold_off + m]
_LOGGER.debug(
    "[milp] hold reward total=%.4f  avg_hold_kwh=%.3f",
    float(np.sum(-c_obj[hold_off : hold_off + m] * hold_sol)),
    float(np.mean(hold_sol)),
)
```

### Files to change

| File | Change |
|---|---|
| `planner/milp_optimizer.py` | Add `hold[t]` variable, objective reward, SoC-coupling inequality, log decode |
| `tests/planner/test_milp_optimizer.py` | Add test: moderate-price slot should not discharge when a high-price slot exists later |

### Acceptance criteria

- [ ] `n_vars` increases from `5*m` to `6*m` with correct offsets
- [ ] Energy balance `A_eq` is unchanged (hold has zero coupling)
- [ ] SoC bounds are not violated by `hold[t]` values
- [ ] A slot with `p_imp[t] = 0.20` followed by `p_imp[t+4] = 0.80` results in `ed[t] ≈ 0` (hold instead of discharge at 0.20)
- [ ] `hold_reward` scales with `avg_import_price` so it works across DKK and EUR without hardcoding
- [ ] Solver time stays under `_SOLVER_TIME_LIMIT_S` with 192 slots and `6*m` variables

### Risk

Low. `hold[t]` has zero coefficient in all equality and most inequality constraints. It only adds `m` new variable bounds and `m` new SoC-coupling rows to `A_ub`. For `m=192` (768 → 1152 variables, 384 → 576 rows) HiGHS solves in well under 1 second. If the hold reward is tuned too high it can suppress profitable discharge — a `hold_reward` of `0.001 * avg_import_price` gives a reward of ~0.0002 DKK/kWh, which is two orders of magnitude below a typical arbitrage spread of 0.20–0.80 DKK/kWh.

woopstar:
## 🔍 MILP Code Review — 3 Remaining Bugs

After reviewing `milp_optimizer.py` in this PR, three issues remain that create inconsistency between the MILP objective and the selector's cost function. None of them cause wrong hardware commands, but they will cause the MILP candidate to score differently under the selector than it should — potentially letting a weaker heuristic candidate win.

---

### Bug A — Cycle cost is double-counted in the MILP objective

**File:** `milp_optimizer.py`
**Lines:** `c_obj[ec_off + t]` and `c_obj[ed_off + t]`

**Problem:**
`cycle_cost_per_kwh` is added to *both* the charge and discharge objective coefficients:

```python
c_obj[ec_off + t] = (cycle_cost_per_kwh + charge_loss * p_imp[t]) * discount
c_obj[ed_off + t] = (cycle_cost_per_kwh + discharge_loss * p_imp[t]) * discount
```

But in `cost_function.py` the selector counts:
```python
throughput_kwh = max(slot.batteries_charged, slot.batteries_discharged)
cycle_cost_total += throughput_kwh * cycle_cost_kwh  # once per slot, not twice
```

One full round-trip = one charge slot + one discharge slot. The MILP charges `cycle_cost_per_kwh` twice (once at charge, once at discharge), so the LP overstates cycle wear by ~2×. This makes the MILP avoid cycling more aggressively than the selector expects, and the resulting plan will show a lower LP objective than the selector's `total_cost` reflects.

**Proposed fix:**
```python
c_obj[ec_off + t] = (cycle_cost_per_kwh / 2 + charge_loss * p_imp[t]) * discount
c_obj[ed_off + t] = (cycle_cost_per_kwh / 2 + discharge_loss * p_imp[t]) * discount
```

**Acceptance criteria:**
- [ ] A unit test with `cycle_cost_per_kwh = 0.10` verifies that the sum of `c_obj[ec] + c_obj[ed]` for a single slot equals `cycle_cost_per_kwh` (not `2 × cycle_cost_per_kwh`)
- [ ] `score_plan(milp_slots)` produces a `total_cost` that is within 1% of the LP objective value `result.fun` on a reference 4-slot test case
- [ ] Existing 1834 tests still pass

---

### Bug B — `hold[t]` variable does not prevent early discharge

**File:** `milp_optimizer.py`
**Lines:** `hold_off`, `A_ub[2 * m + t]`, `c_obj[hold_off + t]`

**Problem:**
`hold[t]` is rewarded for being large (`−hold_reward * discount`), but it is only bounded by `soc[t-1]` — it is **not** coupled to `ed[t]`. The LP can simultaneously set `hold[t] = soc[t-1]` (collect the reward) and `ed[t] = max_dis` (discharge the same energy) in the same slot. The two variables are independent in the energy balance (`hold[t]` has zero coefficient in `A_eq`), so the constraint does not prevent this.

The underlying problem `hold` is trying to solve — preventing the LP from draining the battery before high-price slots — is better addressed with an explicit **terminal SoC opportunity cost** term, consistent with `cost_function.py`'s `terminal_soc_value`.

**Proposed fix:**

Replace the `hold[t]` variable with a terminal SoC credit. The terminal SoC is:
```
soc_terminal = current_kwh + Σ_t(ec[t] − ed[t])
```
which is already linear in `ec` and `ed`. Add a credit for high terminal SoC using `replacement_price_per_kwh`:

```python
# Terminal SoC credit — energy left in battery at end of horizon
# avoids importing at the next discharge window price.
# Adds a negative (reward) coefficient to ec[t] and a positive (cost)
# coefficient to ed[t] for all slots, scaled by discount at horizon end.
if replacement_price is not None and replacement_price > 1e-9:
    terminal_discount = time_discount_rate ** (total_hours_ahead)
    for t in range(m):
        c_obj[ec_off + t] -= replacement_price * terminal_discount
        c_obj[ed_off + t] += replacement_price * terminal_discount
```

This mirrors `terminal_soc_value` in `cost_function.py` exactly and removes the need for `hold[t]` entirely, reducing `n_vars` from `6*m` back to `5*m`.

**Acceptance criteria:**
- [ ] A test with a 6-slot horizon (cheap → hold → expensive) verifies the LP discharges in the expensive slot, not the cheap one, when `replacement_price` matches the expensive slot price
- [ ] When `replacement_price = None`, behaviour is unchanged (no terminal term)
- [ ] `n_vars` is documented in the module docstring and matches the actual variable layout
- [ ] Existing tests pass

---

### Bug C — Mutual exclusion is not guaranteed at negative import prices

**File:** `milp_optimizer.py`
**Lines:** Post-processing block starting with `if ec_kwh > _MIN_ACTION_KWH and ed_kwh > _MIN_ACTION_KWH`

**Problem:**
The docstring states mutual exclusion holds because "the mutual-exclusion constraint together with the per-slot energy caps already prevents simultaneous charge + discharge in the optimal solution." This is only true when `p_imp[t] > 0`. When `p_imp[t] ≤ 0` (negative electricity prices, which occur regularly on the Danish spot market), simultaneously charging and discharging has zero or negative cost in the LP objective, so HiGHS may freely assign `ec[t] > 0` and `ed[t] > 0` in the same slot.

The post-hoc price comparison resolves this but can choose the wrong action when `p_imp ≤ p_exp` at negative prices (the condition `p_imp_t <= p_exp_t` selects charge, which may not be correct when both prices are negative).

**Proposed fix:**

Add a normalised mutual-exclusion inequality to `A_ub` that prevents simultaneous charge + discharge without requiring binary variables:

```python
# Mutual exclusion: ec[t]/max_charge + ed[t]/max_dis <= 1
# This is a valid LP relaxation: in the optimal solution, one of
# ec[t] or ed[t] will be zero because the objective always prefers
# a pure action over a mixed one.
mutex_rows = m
A_mutex = np.zeros((mutex_rows, n_vars))
b_mutex = np.ones(mutex_rows)
for t in range(m):
    A_mutex[t, ec_off + t] = 1.0 / max_charge_per_slot
    A_mutex[t, ed_off + t] = 1.0 / max_dis
A_ub = np.vstack([A_ub, A_mutex])
b_ub = np.concatenate([b_ub, b_mutex])
```

This guarantees mutual exclusion at the solver level and removes the need for post-hoc disambiguation.

**Acceptance criteria:**
- [ ] A test with `p_imp = -0.05` (negative price) verifies that the LP solution has `ec[t] = 0` OR `ed[t] = 0` for every slot (never both `> _MIN_ACTION_KWH`)
- [ ] The post-hoc disambiguation block is removed or kept only as a defensive assertion (`assert not (ec_kwh > ... and ed_kwh > ...)`)
- [ ] Solver feasibility is maintained across the existing test suite
- [ ] Performance: solve time stays under 2.0s with the additional `m` constraint rows

---

### Priority

| Bug | Risk | Effort |
|-----|------|--------|
| A — Cycle cost 2× | Medium — MILP avoids cycling too much | 2 lines |
| C — Negative price mutex | Low–Medium — only triggers at negative DK spot prices | ~10 lines |
| B — `hold[t]` terminal SoC | Medium — LP drains battery suboptimally | ~20 lines + remove hold variable |

Suggested order: **A → C → B**.

woopstar:
## 🔍 Candidate Review — 4 Bugs in `aggressive` and `soc_plan_*` candidates

After reviewing `candidate_generator.py` in this PR, four bugs were found across the `aggressive` and `soc_plan_*` candidates.

---

### Bug D — `aggressive`: charge slots can overlap discharge slots when baseline has no discharge windows

**File:** `candidate_generator.py`
**Function:** `_apply_aggressive_strategy`

**Problem:**

The charge-slot boundary is derived from baseline discharge windows **before** aggressive assigns its own discharge slots:

```python
discharge_starts = sorted(
    s.start for s in future if s.recommendation in _DISCHARGE_RECS
)
earliest_discharge_start = discharge_starts[0] if discharge_starts else None
```

When the baseline has no existing discharge windows, `earliest_discharge_start = None` — no boundary is applied. Aggressive then picks the M most expensive slots as discharge targets, but charge slots have already been selected without knowing where those discharge slots will land. The result: charge slots can be scheduled **after** the discharge slots they are meant to serve, defeating the entire arbitrage intent.

**Proposed fix:**

Derive the charge boundary from the **prospective** discharge slots, not the baseline ones:

```python
# Identify prospective discharge slots first
discharge_candidates = sorted(
    (s for s in future if s.recommendation not in _CHARGE_RECS),
    key=lambda s: (-s.price.import_price, s.start),
)
prospective_discharge_starts = sorted(
    s.start for s in discharge_candidates[:aggressive_discharge_slots]
)
earliest_discharge_start = (
    prospective_discharge_starts[0] if prospective_discharge_starts else None
)
# Then select charge slots using the corrected boundary
```

**Acceptance criteria:**
- [ ] A test with zero baseline discharge windows verifies that every charge slot ends **before** the first aggressive discharge slot starts
- [ ] A test with an existing baseline discharge window verifies the boundary is still respected
- [ ] `total_charge_kwh ≤ usable_kwh` for the aggressive candidate in all tests
- [ ] Existing tests pass

---

### Bug E — `soc_plan_100` and `soc_plan_125` become identical when `total_needed > usable_kwh`

**File:** `candidate_generator.py`
**Function:** `_apply_soc_plan`

**Problem:**

Both candidates hit the same cap:

```python
charge_target = min(charge_needed * charge_fraction, usable_kwh - current_kwh)
```

When `charge_needed` already exceeds `usable_kwh - current_kwh` (e.g. discharge windows demand more energy than the battery can hold), fractions of 1.00 and 1.25 both resolve to `usable_kwh - current_kwh`. The two candidates produce identical slot schedules, wasting one of the 6 candidate slots.

**Proposed fix:**

Deduplicate candidates before appending, or replace `soc_plan_125` with a more useful candidate in saturated cases — e.g. `soc_plan_conservative` that charges to 80% of `usable_kwh` regardless of demand:

```python
if abs(charge_target - prev_charge_target) < 0.05:
    continue  # skip near-identical candidate
prev_charge_target = charge_target
```

**Acceptance criteria:**
- [ ] When `total_needed_kwh > usable_kwh`, no two `soc_plan_*` candidates have the same `charge_target` (within 0.05 kWh tolerance)
- [ ] Total candidate count remains at most 6 `soc_plan_*` entries
- [ ] Existing tests pass

---

### Bug F — `soc_plan_*`: solar slots sorted by import price instead of PV surplus size

**File:** `candidate_generator.py`
**Function:** `_apply_soc_plan`, Step 4

**Problem:**

```python
for slot in sorted(
    (s for s in future if s.recommendation is None),
    key=lambda x: (x.price.import_price, x.start),  # ← wrong key for solar
):
    if slot.estimated_net_consumption < 0.0:  # PV surplus
```

Solar energy is free — import price is irrelevant. Sorting by import price means a slot with 0.01 kWh PV surplus at a low import price is selected over a slot with 2.0 kWh PV surplus at a slightly higher import price. The largest available PV surplus should always be consumed first.

**Proposed fix:**

```python
for slot in sorted(
    (
        s for s in future
        if s.recommendation is None
        and s.estimated_net_consumption is not None
        and s.estimated_net_consumption < 0.0
    ),
    key=lambda x: (x.estimated_net_consumption, x.start),  # most negative = most PV
):
```

This also removes the need for the inner `if` check on `estimated_net_consumption`.

**Acceptance criteria:**
- [ ] A test with two PV-surplus slots (0.5 kWh and 2.0 kWh) verifies the 2.0 kWh slot is selected first regardless of their import prices
- [ ] Total solar energy charged equals `min(sum_of_surpluses, charge_target)` not a subset of it
- [ ] Existing tests pass

---

### Bug G — `soc_plan_*`: `charge_target` does not account for charge efficiency

**File:** `candidate_generator.py`
**Function:** `_apply_soc_plan`

**Problem:**

`charge_needed` is the battery-side energy requirement (already corrected for `discharge_eff`). But `charge_target` is used directly as the grid-import target:

```python
charge_target = charge_needed * charge_fraction
```

To store `charge_needed` kWh in the battery, the grid must supply `charge_needed / charge_eff` kWh (e.g. at 97% efficiency: `charge_needed / 0.97`). `_apply_soc_plan` does not receive `charge_efficiency_pct` and cannot make this correction.

The result: all 6 `soc_plan_*` candidates systematically under-plan grid import by ~3%, leaving the battery ~3% lower than intended at the start of the discharge window.

**Proposed fix:**

Add `charge_efficiency_pct` as a parameter and apply it to `charge_target`:

```python
def _apply_soc_plan(
    ...,
    charge_efficiency_pct: float = 97.0,
    discharge_efficiency_pct: float = 97.0,
) -> None:
    charge_eff = max(min(charge_efficiency_pct, 100.0), 1.0) / 100.0
    ...
    charge_target = (charge_needed * charge_fraction) / charge_eff
    charge_target = min(charge_target, usable_kwh - current_kwh)
```

And pass it through from `generate_candidates`:

```python
_apply_soc_plan(
    ...,
    charge_efficiency_pct=inp.battery_charge_efficiency_pct,
    discharge_efficiency_pct=inp.battery_discharge_efficiency_pct,
)
```

**Acceptance criteria:**
- [ ] At `charge_efficiency_pct=97`, `charge_target` for `soc_plan_100` equals `charge_needed / 0.97` (before the `usable_kwh` cap)
- [ ] After the SoC simulation runs on the `soc_plan_100` candidate, `estimated_battery_soc` at the start of the first discharge window is within 1% of the intended SoC target
- [ ] Existing tests pass

---

### Priority

| Bug | Severity | Effort |
|-----|----------|--------|
| D — aggressive charge/discharge ordering | Medium — arbitrage candidate broken when no baseline discharge windows | ~10 lines |
| F — solar sort key | Low — suboptimal solar capture | 2 lines |
| G — charge efficiency missing | Low — systematic 3% under-charge | ~5 lines + 1 param |
| E — duplicate soc_plan candidates | Low — wastes one candidate slot | ~5 lines |

Suggested order: **D → G → F → E**.

woopstar:
## 🔍 Code Review — 4 additional bugs in `candidate_selector.py` and `milp_optimizer.py`

After reviewing `candidate_selector.py` and `milp_optimizer.py` the following bugs were found.

---

### Bug H — `candidate_selector.py`: `_cost` set as dynamic attribute on a dataclass

**File:** `candidate_selector.py`
**Line:** `candidate._cost = score_plan(...)`

**Problem:**

`CandidatePlan` is a `@dataclass` with no `_cost` field. Setting it dynamically works in CPython today, but the `# type: ignore[attr-defined]` comments throughout the selector reveal the code knows this is wrong. If anyone adds `__slots__` to `CandidatePlan` — a natural memory optimisation for a class that is instantiated once per candidate per planner run — every `candidate._cost` assignment raises `AttributeError` at runtime with no warning at import time.

**Proposed fix:**

Add `_cost` as a proper field to `CandidatePlan` in `candidate_generator.py`:

```python
from dataclasses import dataclass, field
from custom_components.hsem.planner.cost_function import PlanCostBreakdown

@dataclass
class CandidatePlan:
    name: str
    slots: list[PlannedSlot]
    is_valid: bool = True
    rejection_reason: str = ""
    _cost: PlanCostBreakdown | None = field(default=None, repr=False, compare=False)
```

Remove all `# type: ignore[attr-defined]` comments from `candidate_selector.py` once this is in place.

**Acceptance criteria:**
- [ ] `CandidatePlan` has a `_cost` field declared in the dataclass body
- [ ] No `# type: ignore[attr-defined]` comments remain in `candidate_selector.py`
- [ ] Adding `__slots__ = True` to `CandidatePlan` does not break any test
- [ ] Existing tests pass

---

### Bug I — `candidate_selector.py`: `ForceBatteriesDischarge` excluded from `replacement_price_from_next_discharge`

**File:** `candidate_selector.py`
**Function:** `replacement_price_from_next_discharge`

**Problem:**

```python
slot.recommendation == Recommendations.BatteriesDischargeMode.value
```

`_DISCHARGE_RECS` in `candidate_generator.py` contains both `BatteriesDischargeMode` **and** `ForceBatteriesDischarge`. The `aggressive` candidate uses `BatteriesDischargeMode`, but the baseline and other candidates can carry `ForceBatteriesDischarge` slots. When the winning candidate's discharge slots are all `ForceBatteriesDischarge`, `replacement_price_from_next_discharge` finds zero future discharge slots and returns `None`. The terminal SoC value in the cost function is then disabled, making the score incomparable across candidates that do and do not use `ForceBatteriesDischarge`.

**Proposed fix:**

```python
_REPLACEMENT_PRICE_DISCHARGE_RECS = frozenset({
    Recommendations.BatteriesDischargeMode.value,
    Recommendations.ForceBatteriesDischarge.value,
})

# In replacement_price_from_next_discharge:
future_discharge = sorted(
    [
        slot
        for slot in slots
        if (
            slot.recommendation in _REPLACEMENT_PRICE_DISCHARGE_RECS
            and as_tz(slot.start, now.tzinfo) > now
            and not math.isnan(slot.price.import_price)
        )
    ],
    key=lambda s: as_tz(s.start, now.tzinfo),
)
```

**Acceptance criteria:**
- [ ] A test with only `ForceBatteriesDischarge` slots returns a non-`None` replacement price
- [ ] A test with mixed `BatteriesDischargeMode` and `ForceBatteriesDischarge` slots returns the correct average of the top-N prices across both types
- [ ] Existing tests pass

---

### Bug J — `milp_optimizer.py`: simultaneous charge+discharge resolution ignores efficiencies

**File:** `milp_optimizer.py`
**Function:** `solve_milp`

**Problem:**

```python
if p_imp_t <= p_exp_t:
    ed_kwh = 0.0  # prefer charge
else:
    ec_kwh = 0.0  # prefer discharge
```

When import and export prices are close (e.g. `p_imp=1.50`, `p_exp=1.45`) and round-trip efficiency is 94% (97% × 97%), the net arbitrage profit is negative: charging costs `1.50 / 0.97 = 1.546` and discharging yields `1.45 × 0.97 = 1.407`. The battery loses money on the round trip. The current resolution code always picks one action over idle, so MILP can produce a plan that is more expensive than `no_action` for these edge-case prices.

**Proposed fix:**

Resolve to idle when the round-trip is not profitable:

```python
if ec_kwh > _MIN_ACTION_KWH and ed_kwh > _MIN_ACTION_KWH:
    net_charge_profit = (
        p_exp[lp_t] * discharge_eff
        - p_imp[lp_t] / charge_eff
        - 2.0 * cycle_cost_per_kwh
    )
    if net_charge_profit > 0:
        # Charging is net profitable — keep charge, drop discharge
        ed_kwh = 0.0
    else:
        # Round trip is not profitable — drop both (idle)
        ec_kwh = 0.0
        ed_kwh = 0.0
```

**Acceptance criteria:**
- [ ] A test with `p_imp=1.50`, `p_exp=1.45`, `charge_efficiency_pct=97`, `discharge_efficiency_pct=97` resolves simultaneous charge+discharge to idle (both zero)
- [ ] A test with `p_imp=1.00`, `p_exp=1.80` resolves to charge-only
- [ ] MILP candidate total cost is never higher than `no_action` total cost in a parameterised test across 20 random price scenarios
- [ ] Existing tests pass

---

### Bug K — `milp_optimizer.py`: `hold_reward` can produce `nan` via `np.mean` on empty array

**File:** `milp_optimizer.py`
**Function:** `solve_milp`

**Problem:**

```python
hold_reward = (
    0.001 * float(np.mean(p_imp[p_imp > 1e-9])) if np.any(p_imp > 1e-9) else 1e-4
)
```

The guard `np.any(p_imp > 1e-9)` can evaluate to `True` while `p_imp[p_imp > 1e-9]` is still an empty array when prices are exactly `1e-9` due to floating-point comparison ordering between `np.any` and boolean indexing in edge cases. `np.mean` on an empty array returns `nan` with a `RuntimeWarning`. A `nan` in the objective vector `c_obj` causes HiGHS to return a failed result, the function returns `None`, and the MILP candidate is silently dropped with only a `debug` log — no indication to the user that the solver was fed invalid data.

**Proposed fix:**

```python
positive_prices = p_imp[p_imp > 1e-9]
hold_reward = (
    0.001 * float(np.mean(positive_prices))
    if len(positive_prices) > 0
    else 1e-4
)
```

Also add a NaN-guard on `c_obj` before calling `linprog` to surface this class of issue explicitly:

```python
if np.any(np.isnan(c_obj)):
    _LOGGER.warning("[milp] NaN detected in objective vector — aborting solver")
    return None
```

**Acceptance criteria:**
- [ ] A test with all import prices set to exactly `1e-9` does not produce a `RuntimeWarning` and returns either a valid slot list or `None` (not a crash)
- [ ] A test with all import prices set to `0.0` returns `None` gracefully (no solver exception)
- [ ] `c_obj` NaN guard is present and triggers a `warning`-level log
- [ ] Existing tests pass

---

### Priority

| Bug | File | Severity | Effort |
|-----|------|----------|--------|
| I — `ForceBatteriesDischarge` missing from replacement price | `candidate_selector.py` | Medium — terminal SoC value disabled for some candidates | ~5 lines |
| J — simultaneous charge/discharge ignores efficiencies | `milp_optimizer.py` | Medium — MILP can be worse than no_action | ~8 lines |
| H — `_cost` dynamic attribute on dataclass | `candidate_selector.py` | Medium — latent crash if `__slots__` added | ~3 lines |
| K — `hold_reward` NaN edge case | `milp_optimizer.py` | Low — MILP silently dropped on degenerate prices | 2 lines |

Suggested order: **I → J → H → K**.

woopstar:
## Code review — issues to fix before merge

After reading the actual diff, here are the concrete issues that need to be addressed:

---

### 1. `time_discount_rate` default inconsistency

`PlannerInput.time_discount_rate` defaults to `1.0` (no discount), but `coordinator.py` hardcodes `0.995`. This means:
- Unit tests using `PlannerInput()` directly get **no discount** (different behaviour than production)
- The discount is silently always active in production, but never tested as such

**Fix:** Either set the default in `PlannerInput` to `0.995` to match production, or add a test that constructs `PlannerInput` via `_build_planner_input()` and verifies the discount is applied.

---

### 2. `concentrate_discharge_on_expensive_slots` ignores `ForceBatteriesDischarge`

The function only looks for `Recommendations.BatteriesDischargeMode.value` but `_DISCHARGE_RECS` also includes `ForceBatteriesDischarge`. This means forced discharge slots are invisible to the concentration logic and won't be cleared even if the battery can't serve them.

**Fix:** Replace the hardcoded `== Recommendations.BatteriesDischargeMode.value` filter with `in _DISCHARGE_RECS` (same as was done in `replacement_price_from_next_discharge` for Bug I).

---

### 3. `_compute_replacement_price` in MILP still uses hardcoded `BatteriesDischargeMode`

`replacement_price_from_next_discharge` was fixed to use `_DISCHARGE_RECS` (Bug I), but the equivalent logic inside the MILP optimizer (`_compute_replacement_price` or similar) still only checks `BatteriesDischargeMode.value`. This is an inconsistency — the MILP solver's terminal-SoC valuation will miss forced-discharge slots.

**Fix:** Apply the same `in _DISCHARGE_RECS` fix there too.

---

### 4. Merge conflicts

The PR is currently marked `dirty` (merge conflicts with `main`). These need to be resolved before merge.

---

### Minor / non-blocking

- The `2×` factor in `_resolve_cycle_cost` is correct and well-documented, but the docstring in `calculate_recommended_threshold` (called from `working_mode_sensor.py`, `flows/batteries_excess_export.py`, and `engine.py`) should be updated to reflect that `conversion_loss` is no longer a parameter, so callers don't pass `0.0` by habit.
- `commit_msg.txt` was deleted — good. But the leftover `# type: ignore[attr-defined]` comments referenced in the selector were cleaned up, which is correct now that `_cost` is a proper dataclass field.

---

**Summary:** Issues 1–3 are the functional bugs. Issue 4 (merge conflicts) is a hard blocker. Everything else looks good.

woopstar:
## Claude Prompt — Implement MILP Option 1 + Update Refactor Issues

This PR introduces new functions into `engine.py`, `charge_scheduler.py`, and `milp_optimizer.py`. Before merging, the following tasks must be done in one coordinated step to avoid the refactor issues (#441, #442) having to re-process this PR's new code after merge.

---

### Task 1 — Implement MILP Option 1 (promote MILP to primary optimizer)

The current state after this PR: MILP is wired as a 7th candidate. The goal is to promote it to the **primary optimizer** and demote the rule-based baseline to fallback-only.

```
You are working in the HSEM repository on branch for PR #425.

Context:
- planner/milp_optimizer.py exists and returns a recommendation list per slot
- planner/candidate_generator.py generates 7 candidates including one named "milp"
- planner/candidate_selector.py scores and selects the best candidate
- planner/engine.py wires everything together via run_planner()

Goal: Promote the MILP solution to the primary optimizer path.

Step 1: In planner/engine.py, change the candidate wiring so that:
  A) The MILP optimizer runs first
  B) If solve_milp() returns a valid result, it is used directly as the plan — skip the
     candidate generation loop entirely
  C) If solve_milp() returns None (infeasible or timeout), fall back to the existing
     7-candidate heuristic pipeline unchanged

The fallback chain must be:
  solve_milp() → success → apply MILP plan → score → done
  solve_milp() → None    → generate_candidates() → select_best_candidate() → done

Step 2: In planner/candidate_generator.py, remove the "milp" candidate from
generate_candidates(). It should no longer be part of the 7-candidate set because
the MILP now runs before candidate generation, not inside it.

Step 3: In planner/milp_optimizer.py, increase the solver timeout from 2.0s to 5.0s
for the primary path (not the fallback). Add a comment explaining why.

Step 4: In planner/candidate_selector.py, add a guard so that if the input candidate
list is empty (possible if MILP was used and no heuristic candidates were generated),
the function returns None gracefully instead of raising an IndexError.

Step 5: Update the planner output dataclass (PlannerOutput or equivalent) to include
a field: optimizer_used: str  — set to "milp" or "heuristic" depending on which path
was taken. This allows the HA sensor to report which optimizer ran.

Step 6: Add or update tests in tests/planner/ to cover:
  - MILP primary path: assert optimizer_used == "milp" on a solvable input
  - Fallback path: mock solve_milp to return None, assert optimizer_used == "heuristic"
  - Infeasible guard: assert no exception when MILP is infeasible

Do not change the MILP formulation itself — only the wiring and fallback logic.
Run the full test suite after each step. All existing tests must pass.
```

---

### Task 2 — Update refactor issue #441 (engine.py split) to account for this PR's additions

Issue #441 plans to split `engine.py` into `engine_core`, `engine_explanation`, and `engine_hardware`. This PR adds new logic to `engine.py` (MILP wiring, `optimizer_used` field, fallback chain). The Claude prompt in #441 must be updated to include these additions.

```
Update the Claude prompt in issue #441 (refactor: split engine.py into engine_core,
engine_explanation, engine_hardware) to account for the changes introduced in PR #425.

Changes from PR #425 that affect the split:
1. A new MILP primary path was added to run_planner() — this belongs in engine_core.py
2. The fallback chain (solve_milp → heuristics) is part of the core planning flow,
   not explanation or hardware — ensure it lands in engine_core.py
3. optimizer_used field on PlannerOutput is set in engine_core.py

Update the "Step 1: Analyse engine.py" instruction in issue #441 to explicitly list
these new functions/paths as belonging to category A (Core planning flow):
  - solve_milp() call and result check
  - fallback chain logic
  - optimizer_used assignment

Do not change any other part of the issue.
Update the issue body with the revised Claude prompt.
```

---

### Task 3 — Update refactor issue #442 (charge_scheduler.py split) to account for this PR's additions

Issue #442 plans to split `charge_scheduler.py` into `charge_scheduler.py` (charge only) and `discharge_scheduler.py` (discharge only). This PR adds `concentrate_discharge_on_expensive_slots()` to `charge_scheduler.py`. The Claude prompt in #442 already lists this function, but must be checked for completeness.

```
Update the Claude prompt in issue #442 (refactor: split charge_scheduler.py into charge
and discharge modules) to account for the changes introduced in PR #425.

Changes from PR #425 that affect the split:
1. concentrate_discharge_on_expensive_slots() was added to charge_scheduler.py
   — this is a discharge-adjacent function and must go into discharge_scheduler.py
2. The function uses _DISCHARGE_RECS internally — after issue #439 is done,
   it should import DISCHARGE_RECS from utils/recommendations.py instead

Verify that the existing Claude prompt in #442 already categorises
concentrate_discharge_on_expensive_slots as discharge-related (category B).
If it does, add an explicit note:
  "Note: concentrate_discharge_on_expensive_slots was added in PR #425.
   It is confirmed as discharge-related and must move to discharge_scheduler.py."

If it does not already list it, add it to Step 2 category B.

Do not change any other part of the issue.
Update the issue body with the revised Claude prompt.
```

---

### Notes

- Tasks 2 and 3 are issue-body updates, not code changes — they prevent the refactor issues from having stale prompts after this PR merges.
- Task 1 is the primary code change and must be done on the PR #425 branch before merge.
- Run `pytest` after Task 1 and confirm all tests pass before updating the issues.

PR #627, opened 2026-07-07
fix(recommendation-resolver): prevent ev_smart_charging label when planner says stop

woopstar:
## Problem

When the planner sets `ev_charger_calculated_power` to 0 (stop charging command — e.g. target SoC reached, expensive grid power, no surplus PV), the `recommendation_resolver` was overriding the recommendation to `ev_smart_charging` just because the EV was physically charging. This created an inconsistency: `recommendation: ev_smart_charging` with `ev_charger_calculated_power: 0`.

## Root Cause

`ev_charger_calculated_power` is HSEM's **command** to the charger, not a reflection of what the charger is doing. When the planner sets it to 0, that means "stop charging". The resolver was ignoring this and overriding based solely on `live.ev.is_charging`.

## Fix

The resolver now only overrides to `ev_smart_charging` when **both** conditions are true:
1. The EV is physically charging (`live.ev.is_charging`)
2. The planner allocated EV load for this slot (`ev_charger_calculated_power > 0` OR `ev_total_planned_load_kwh > 0`)

If the planner set power to 0 (stop command), the resolver respects that decision and keeps the planner's original recommendation.

## Files Changed

- `custom_components/hsem/custom_sensors/recommendation_resolver.py` — Added `planner_allocated_ev` guard before EV override
- `tests/sensors/test_recommendation_resolver.py` — Added tests for new guard condition, restored `TestDataStateSync` tests
- `tests/planner/test_slot_ownership.py` — Updated existing test to set `ev_charger_calculated_power > 0`

### DIFF
--- a/custom_components/hsem/custom_sensors/recommendation_resolver.py
+++ b/custom_components/hsem/custom_sensors/recommendation_resolver.py
@@ -53,20 +53,24 @@ def resolve_current_recommendation(
     if rec.recommendation == Recommendations.BatteriesChargeGrid.value:
         return
 
-    # 3. Any EV is actively charging → override with EV smart charging.
+    # 3. Any EV is actively charging AND the planner allocated EV load for
+    #    this slot → override with EV smart charging.
     #
-    # Guard: only apply the live override when the planner already injected EV
-    # planned load for this slot (ev_planned_load_kwh > 0) OR when the planner
-    # had no EV data at all (ev_planned_load_kwh == 0 because the feature was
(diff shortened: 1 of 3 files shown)


woopstar:
Pushed a small follow-up commit (`16c4235`): updated the two EV smart-charging tests to set `ev_charger_calculated_power` / `ev_second_charger_calculated_power` so they match the new planner-allocated guard, and dropped the now-unused `Any` import in the test file.

PR #673, opened 2026-07-17
fix(migration): migrate entity unique_ids in place on v1->v2

heinoskov:
Problem

v6.0.0 (#523) prefixed every entity unique_id with the config entry id
(hsem_workingmode_sensor → hsem_<entry_id>_workingmode_sensor), but the
async_migrate_entry v1→v2 migration only migrates the config data — it
never migrates the entity registry.

So on the first load after upgrading from v5.x, Home Assistant does not
recognise the new unique_ids as belonging to the existing entities. Each
entity is re-created fresh:

the old entity is orphaned ("no longer provided by the hsem integration"), and
the live entity is registered with a _2 suffix (sensor.hsem_..._2), because
the original entity_id is still taken by the orphan.

This contradicts the release notes' own guarantees:

"Entity IDs are unchanged" — false; live entities become *_2.
"State history is preserved" — the old long-term statistics stay stranded on
the orphan; the live _2 entity starts from zero.

Any automations/dashboards/templates referencing sensor.hsem_* then point at the
dead orphan.

Reproduced on a real v5.1.0 → v6.0.0 upgrade: 243 orphaned + 170 _2 entities.

Fix

Add the missing entity-registry migration to async_migrate_entry: rename each
existing entry's unique_id from hsem_<key> to hsem_<entry_id>_<key> in
place via entity_registry.async_migrate_entries. The same registry entry is
reused, so entity_id and long-term statistics are preserved and no _2
duplicates are created. The guard skips already-prefixed ids → idempotent and
safe for genuinely new v6 entities.

Verification

Applied to a live v5.1.0 → v6.0.0 upgrade:

0 _2 duplicates, 0 orphans
migrated entity keeps its entity_id; unique_id is the new hsem_<entry_id>_…
form; history intact
HSEM healthy afterwards (workingmode live, read_only honoured, apply_status
ok, data_quality complete)

Adds tests/test_config_flow_migrate_unique_ids.py covering the in-place remap and
idempotency.

### DIFF
--- a/custom_components/hsem/config_flow.py
+++ b/custom_components/hsem/config_flow.py
@@ -8,6 +8,7 @@ import voluptuous as vol
 from homeassistant import config_entries
 from homeassistant.config_entries import ConfigEntry, ConfigFlowResult
 from homeassistant.core import HomeAssistant, callback
+from homeassistant.helpers import entity_registry as er
 
 from custom_components.hsem.const import DEFAULT_CONFIG_VALUES, DOMAIN, NAME
 from custom_components.hsem.flows.batteries_excess_export import (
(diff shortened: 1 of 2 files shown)


woopstar:
can you fix the lint check issue. Sonarcloud is ok to fail. Seems to be some other problem

heinoskov:
> can you fix the lint check issue. Sonarcloud is ok to fail. Seems to be some other problem

working on it.

heinoskov:
> can you fix the lint check issue. Sonarcloud is ok to fail. Seems to be some other problem

done :-)

PR #716, opened 2026-08-05
fix(sensor): pass hass to IntegrationSensor/UtilityMeterSensor only when accepted

proscar87:
## What broke

`home-assistant/core` merged two PRs on 2026-07-30 that removed the `hass` parameter from two helper-entity `__init__` methods, shipping in **HA 2026.8.0** (released 2026-08-05):

- [home-assistant/core#177596](https://github.com/home-assistant/core/pull/177596) — "Do not set a device on YAML integration entities" (merged 2026-07-30T06:49:24Z): `IntegrationSensor.__init__` went from `def __init__(self, hass: HomeAssistant, *, integration_method, name, round_digits, source_entity, unique_id, unit_prefix, unit_time, max_sub_interval) -> None:` to `def __init__(self, *, integration_method, name, round_digits, source_entity, unique_id, unit_prefix, unit_time, max_sub_interval, device=None) -> None:` — `hass` removed entirely.
- [home-assistant/core#177603](https://github.com/home-assistant/core/pull/177603) — "Do not set a device on YAML utility_meter entities" (merged 2026-07-30T06:46:52Z): same change to `UtilityMeterSensor.__init__`.

**Neither change is listed in HA's official breaking-changes documentation** — both classes are considered internal API by core maintainers, so integrations that build on them got no warning.

`custom_components/hsem/custom_sensors/house_consumption_power_sensor.py` instantiates `HSEMIntegrationSensor` and `HSEMUtilityMeterSensor` with `hass=self.hass` as a keyword argument at both call sites (~line 477 and ~line 540). Both wrapper classes blindly forward `**kwargs` (including `hass`) straight into the parent HA class's `__init__`:

- `custom_components/hsem/custom_sensors/integration_sensor.py` → `IntegrationSensor.__init__(self, *args, **kwargs)`
- `custom_components/hsem/custom_sensors/utility_meter_sensor.py` → `UtilityMeterSensor.__init__(self, *args, **kwargs)`

On HA 2026.8+, this raises `TypeError: __init__() got an unexpected keyword argument 'hass'` at entity-creation time — every derived integral/utility-meter sensor HSEM creates would fail to instantiate on upgrade.

I independently confirmed this by installing `homeassistant==2026.8.0` in an isolated venv and reproducing the exact `TypeError` by calling `IntegrationSensor(hass=..., ...)` directly.

## The fix

Rather than pinning `homeassistant` to an older version, detect at runtime whether the installed parent `__init__` still accepts `hass` via `inspect.signature(...)`, and only forward it when it does. This way the same code works on both pre-2026.8 (which requires `hass`) and 2026.8+ (which rejects it) — no dependency pin change needed.

```python
if "hass" not in inspect.signature(IntegrationSensor.__init__).parameters:
    kwargs.pop("hass", None)
IntegrationSensor.__init__(self, *args, **kwargs)
```

...and the mirror in `utility_meter_sensor.py` checking `UtilityMeterSensor.__init__`'s own signature independently (they happened to change identically in this HA release, but the two checks are not assumed to be coupled).

`house_consumption_power_sensor.py` (the call sites) is untouched — the fix is entirely in the two wrapper `__init__` methods, keeping the diff minimal.

### Precedent

This is the same pattern already used successfully elsewhere for this exact upstream change:

- [jseidl/magic-areas#627](https://github.com/jseidl/magic-areas/pull/627) — "Keep AreaThresholdSensor working on HA 2026.8", same `inspect.signature` technique for `ThresholdSensor`.
- `bramstroker/homeassistant-powercalc` — has used runtime signature detection for HA helper-entity compatibility for years.
- `nathanmarlor/foxess_modbus` — explicit `inspect.signature` usage for the same class of problem.
- `Olen/homeassistant-plant#500` — same fix pattern.

## Testing

**Important existing-test caveat found during this work:** `tests/test_house_consumption_sensor_lifecycle.py` tests `HSEMHouseConsumptionPowerSensor` by monkeypatching `HSEMIntegrationSensor.__init__` and `HSEMUtilityMeterSensor.__init__` themselves with fakes (`patch.object(HSEMIntegrationSensor, "__init__", _fake_integration_init)`). That means it never exercises the real constructor path and would not have caught this bug, nor would it prove this fix works.

Added `tests/test_integration_utility_meter_sensor_init.py` with 6 new tests that do **not** patch the HSEM wrapper `__init__` methods:

- `TestRealConstructor` — constructs the real, unpatched `HSEMIntegrationSensor`/`HSEMUtilityMeterSensor` against whatever `homeassistant` is actually installed.
- `TestSimulatedHa2026_8Signature` — patches only the *parent* HA class's `__init__` (`IntegrationSensor.__init__` / `UtilityMeterSensor.__init__`) with a stand-in mirroring the real 2026.8+ signature (no `hass` param), so the real HSEM wrapper logic (including this fix) is genuinely exercised and the test is deterministic regardless of which `homeassistant` version happens to be pinned at test time. Also includes a backward-compat test confirming `hass` is still forwarded when the parent signature accepts it.

**Discrepancy worth flagging:** `requirements.txt`/`pyproject.toml` currently pin `homeassistant==2026.7.4`, which still accepts `hass` — so this specific `TypeError` does not reproduce today against the pinned test environment (only the `TestSimulatedHa2026_8Signature` tests actually prove the fix works). I independently confirmed the real bug and the real fix by installing `homeassistant==2026.8.0` in a separate throwaway venv, reproducing `TypeError: IntegrationSensor.__init__() got an unexpected keyword argument 'hass'`, and confirming `UtilityMeterSensor.__init__` dropped `hass` in the same way. I deliberately did **not** bump the `homeassistant` pin in this PR — that's outside the minimal-diff scope of this fix and is a separate decision for the maintainer; the runtime-detection fix is correct regardless of which version is installed.

I removed and re-applied the fix locally (`git stash`) to confirm the two `TestSimulatedHa2026_8Signature` "drops_hass" tests genuinely fail with the exact real-world `TypeError` on pre-fix code and pass after the fix.

### Quality gates (per `AGENTS.md`/`CLAUDE.md`)

All run with Python 3.14.2 via `uv sync --all-groups`:

| Gate | Before (baseline, `main`) | After (this branch) |
|---|---|---|
| `./scripts/quality.sh lint` | pass (ruff format + check) | pass |
| `./scripts/quality.sh typing` | pass (mypy, 276 files) | pass (mypy, 277 files) |
| `./scripts/quality.sh quality` | pass (pyright + vulture, 0/0/0) | pass (pyright + vulture, 0/0/0) |
| `./scripts/quality.sh test` | 2384 passed | 2390 passed (6 new) |

No pre-existing failures were encountered; nothing needed to be worked around.

## Disclosure

This PR was implemented with [Claude Code](https://claude.com/claude-code). I have no real Huawei solar hardware and validated this change entirely via unit tests and mocks (including a manual, isolated-venv reproduction of the upstream `TypeError` against `homeassistant==2026.8.0`) — no manual/hardware testing was performed.

### DIFF
--- a/custom_components/hsem/custom_sensors/integration_sensor.py
+++ b/custom_components/hsem/custom_sensors/integration_sensor.py
@@ -6,6 +6,7 @@ metadata so that all derived energy sensors appear on the HSEM device page.
 
 from __future__ import annotations
 
+import inspect
 from typing import Any, override
 
 from homeassistant.components.integration.sensor import IntegrationSensor
@@ -42,6 +43,14 @@ class HSEMIntegrationSensor(IntegrationSensor, HSEMEntity):
             config_entry: The HSEM config entry (required).
             **kwargs: Keyword arguments forwarded to :class:`IntegrationSensor`.
         """
+        # Home Assistant core removed the `hass` parameter from
+        # IntegrationSensor.__init__ in 2026.8 (home-assistant/core#177596)
+        # without listing it as a breaking change, since the class is
+        # considered internal API. Only pass `hass` through when the
+        # installed HA version still accepts it, so this works on both
+        # sides of that change.
+        if "hass" not in inspect.signature(IntegrationSensor.__init__).parameters:
+            kwargs.pop("hass", None)
         IntegrationSensor.__init__(self, *args, **kwargs)
         assert config_entry is not None, (
             "config_entry is required for HSEMIntegrationSensor"
--- a/custom_components/hsem/custom_sensors/utility_meter_sensor.py
+++ b/custom_components/hsem/custom_sensors/utility_meter_sensor.py
@@ -6,6 +6,7 @@ per-hour energy meters appear on the HSEM device page.
 
 from __future__ import annotations
 
+import inspect
 from typing import Any, override
 
 from homeassistant.components.sensor.const import SensorDeviceClass, SensorStateClass
@@ -43,6 +44,14 @@ class HSEMUtilityMeterSensor(UtilityMeterSensor, HSEMEntity):
             config_entry: The HSEM config entry (required).
             **kwargs: Keyword arguments forwarded to :class:`UtilityMeterSensor`.
         """
+        # Home Assistant core removed the `hass` parameter from
+        # UtilityMeterSensor.__init__ in 2026.8 (home-assistant/core#177603)
+        # without listing it as a breaking change, since the class is
+        # considered internal API. Only pass `hass` through when the
+        # installed HA version still accepts it, so this works on both
+        # sides of that change.
+        if "hass" not in inspect.signature(UtilityMeterSensor.__init__).parameters:
+            kwargs.pop("hass", None)
         UtilityMeterSensor.__init__(self, *args, **kwargs)
         assert config_entry is not None, (
             "config_entry is required for HSEMUtilityMeterSensor"
--- a/tests/test_integration_utility_meter_sensor_init.py
+++ b/tests/test_integration_utility_meter_sensor_init.py
@@ -1,0 +1,288 @@
+"""Regression tests for HSEMIntegrationSensor / HSEMUtilityMeterSensor construction
+against both the pre-2026.8 and 2026.8+ Home Assistant core signatures.
+
+home-assistant/core PR #177596 ("Do not set a device on YAML integration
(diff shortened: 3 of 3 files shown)


woopstar:
Released in v6.2.0-beta1

woopstar:
@proscar87 can you confirm working before I release v6.2.0

PR #724, opened 2026-08-09
fix(translations): repair mojibake, swapped interval descriptions, wrong unit labels

breswa:
Spotted these while walking the v6.1.0 options flow page by page (with help from Anthropic's Claude Fable, which also prepared this patch — flagging that for transparency).

Strings-only, no code changes:

- **en.json — 50 double-encoded characters restored.** En/em dashes rendered as `â€"`, `×` as `Ã—`, `°` as `Â°` across the economics, EV, power and ML pages. Fixed via cp1252 round-trip; no residual, file stays valid JSON.
- **da.json — 10 instances**, fixed from context: `×` in the two efficiency formula strings, `é` in the "én måned" validation strings. We're not Danish speakers — worth a native-eye check that these read right.
- **Recommendation Interval Minutes / Length descriptions were swapped** — the "length in minutes of each interval" text sat under Length (the horizon field) and the "number of intervals" text under Minutes. Swapped in all three flow sections.
- **Unit labels:** max charging power said "kilowatts (W)" (the sensor is watts); rated capacity said "(Watt)"/"watts (W)" (the register is watt-hours) — now "(Wh)"/"watt-hours (Wh)". Happy to split these out if you'd rather keep the PR to pure encoding fixes.


woopstar:
Many thanks!

breswa:
Thanks for such a great Git!

PR #760, opened 2026-08-14
fix(applier): verify TOU schedule against Period attributes, not entity state

Ambilights:
## Problem

The TOU write is verified by hashing the desired period list and comparing it against a hash of the TOU entity's **state**. That state is only the *number* of configured periods — the schedule itself lives in the entity's `Period 1`…`Period 10` attributes. The two can never match, so a successful write is always reported as `FAILED`.

```python
desired=generate_hash(str(tou_modes)),                       # ['00:00-00:01/1234567/+']
reader=lambda: generate_hash(str(states.get(tou_entity).state)),   # "1"
```

Confirmed by recomputing the hashes from a live failure:

```
desired = a5e4eea0…  = sha256("['00:00-00:01/1234567/+']")
actual  = 6b86b273…  = sha256("1")
```

Because the failure path returns early, **every genuine TOU period change also blocks the remaining battery writes for that cycle** — working mode, discharge cap, and everything after it.

## Evidence

Observed on a fork of this integration running 6.2.2. The write itself succeeds on the first attempt; only the verification fails:

```
11:10:22.541  Set TOU periods … ['00:00-00:01/1234567/+']
11:10:22.541  TOU entity last_updated → Period 1: 00:00-00:01/1234567/+   ← applied
11:10:43.816  ERROR write-and-verify FAILED after 2 attempt(s)
11:12:20.161  next hardware command finally lands                        ← +96 s
```

`last_changed` on that entity does not move, because the state (`"1"`) is unchanged — only `last_updated` and the attribute do. That is the same reason the verification cannot see the write.

The condition is self-limiting rather than a retry loop: the write gate compares against `live.tou_periods.periods`, which is attribute-derived and therefore correct, so once the hardware matches no further writes are attempted. It fires once per genuine payload change.

## Fix

Verification re-reads the entity and compares the `Period 1`…`Period 10` attributes.

It has to read Home Assistant directly rather than `LiveState` — that snapshot is captured before the write and by definition still holds the old schedule, so verifying against it would fail on every attempt for the same underlying reason.

The Period-attribute parsing already existed in `state_collector.py`; this moves it to `utils/huawei.extract_tou_periods()` so the applier and collector share one definition.

Also corrects the comment claiming "no read-back verification" and "single attempt only", which contradicts the call it documents (`max_retries=2`, with verification). That comment is plausibly what kept this hidden.

## Validation

- 2517 existing tests pass, plus 6 new ones covering attribute extraction and the read-back helper
- `ruff format` / `ruff check`, `mypy`, `pyright` clean on Python 3.14.7
- Running in production on the reporting system: two TOU writes in both directions (placeholder ↔ full-day window), each verified on the first attempt, no ERROR and no aborted cycles

## Unrelated, noted in passing

`_LOGGER.debug("…", "warning")` appears in several places in `applier.py`, passing a level name as a positional `%`-format argument to a message with no placeholders. Not touched here.

### DIFF
--- a/custom_components/hsem/custom_sensors/applier.py
+++ b/custom_components/hsem/custom_sensors/applier.py
@@ -57,6 +57,7 @@ from custom_components.hsem.utils.huawei import (
     async_set_grid_export_power_watt,
     async_set_tou_periods,
     async_stop_forcible_discharge,
+    extract_tou_periods,
 )
 from custom_components.hsem.utils.inverter_verify import (
     ApplyResult,
@@ -727,8 +728,11 @@ async def async_apply_battery_settings(
             )
             return summary
 
-    # TOU periods — no read-back verification (TOU period state is complex JSON;
-    # hash comparison is sufficient; single attempt only).
+    # TOU periods — verified against the entity's live ``Period N`` attributes
+    # (see :func:`_read_tou_periods`).  The gate below uses the pre-write
+    # LiveState snapshot only to decide *whether* a write is needed;
+    # verification must re-read HA, because that snapshot by definition still
+    # holds the old schedule.
     if (
         working_mode == WorkingModes.TimeOfUse.value
         and tou_modes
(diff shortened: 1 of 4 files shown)


woopstar:
Good catch !

Issue #767, opened 2026-08-18
applier writes a 0 W grid feed-in limit below export_electricity_min_price, which disables PV export entirely


breswa:
### Summary

Below `export_electricity_min_price`, the applier sets the inverter's grid feed-in limit to WATT mode at 0 W. That register governs the whole grid connection point, so it does not just stop battery-to-grid export — it stops ALL export, including surplus PV that has nowhere else to go once the battery is full.

### Code

`custom_components/hsem/custom_sensors/applier.py` (current main):

```python
:263   export_pct = 100 if export_price >= min_price else 0
:331   mode = "W" if export_pct == 0 else "%"
```

With `export_electricity_min_price` = 0.22 and a non-peak export price of $0.10, `export_pct` is 0 for most of every day, so the applier writes watt-mode 0 W.

### Why this reads as a bug rather than the intended guard

The setting is about price — "don't sell below this" — and that intent is about the BATTERY, whose energy has a replacement cost. Surplus PV has no replacement cost: once the battery is full, the alternative to exporting at $0.10 is curtailing at $0.00. Blocking the feed-in limit takes the strictly worse option, silently.

#744 (separate PV and battery export thresholds) and #752 (battery export floor) suggest the conflation is already on your radar as a feature gap. What makes this feel like a bug rather than a missing feature is that the PV side is collateral damage with no indication: nothing in HA reports that export has been disabled. The only visible symptom is `sensor.inverter_active_power_control` changing from percent mode to "Limited to 0W", and on our install that entity was disabled by default.

For completeness, `battery_export_min_price` (#752 / `ecd9ae22`, v6.2.2) does not address this — it adds a planner-side floor for intentional battery export and does not change the applier's feed-grid write. I re-checked lines 263 and 331 against main today and both are still present.

### Impact on our install

Six days of zero PV export (2026-08-09 to 2026-08-17) on a 14 kWp / 10 kW AC two-inverter site. Recovered by writing the register back by hand:

```yaml
service: huawei_solar.set_maximum_feed_grid_power
data:
  device_id: <inverter>
  power: 4999
```

which also flips the mode register back to `POWER_LIMITED_GRID_CONNECTION_WATT`. Export resumed immediately, 0 -> 2146 W.

### Honest limits on our evidence

At present we are running `read_only` ON (hsem reports `hardware_writes_blocked`), so I have **not** observed hsem performing the write live. The evidence is:

- the code path above, which is reachable at our prices, and
[...]

PR #928, opened 2026-09-06
fix(sensors): stop PV Curtailment sensor false-positive at configured export cap

### LINKED ISSUE #924: PV curtailment detected but inverter export limit remains at 10 kW (Inverter Apply = Pending)
Hi,

I’m seeing an issue with PV export limiting / curtailment in the current HSEM beta.

When the export electricity price is below my configured minimum export threshold, HSEM correctly detects that PV should be curtailed:

PV Curtailment = curtailed
Hardware Writes = Allowed
Plan Strategy = milp

However, the inverter setting is not actually updated.

The HSEM dashboard keeps showing:

Inverter Apply = Pending

At the same time, the Huawei inverter still shows:

Active Power Control Mode: Limited Power Grid (kW)
Max. grid feed-in power: 10.000 kW

So when the battery is full, excess PV is still exported to the grid even though the export price is below the configured threshold.

I also tested this while starting EV charging to consume some of the PV locally. HSEM still reported PV Curtailment = curtailed, but the inverter export limit remained unchanged at 10 kW.

System:

Huawei SUN2000-10KTL-M1
Huawei battery
10 kW grid export limit
HSEM MILP planner
Home Assistant
Hardware writes enabled

It looks like the planner makes the correct curtailment decision, but the hardware applier does not successfully write the new active power/export limit to the inverter.

I can provide diagnostics, logs and screenshots if needed.

woopstar:
## Summary

- Fixes a false-positive in the **PV Curtailment** diagnostic sensor
  (`sensor.hsem_pv_curtailment_sensor`) reported in #924: it showed
  `curtailed` permanently on any installation with a configured
  `max_grid_export_power_kw` (DNO/grid export cap), even though the inverter
  was simply sitting at its routine, always-applied cap — not being
  actively throttled for price/SoC reasons.

## Root cause

Per issue #767's design (documented in `.github/memories.md`), the applier
(`async_apply_inverter_power_control`) writes the configured grid-export cap
in watts to the inverter for **every non-negative export price** — it only
drops to a hard near-zero block when the export price is genuinely
negative. This is intentional: surplus PV export should continue up to the
DNO cap even at low positive prices, since curtailing free PV production
wastes it for no benefit.

`HSEMPVTailedSensor._is_directly_limited()` treated *any* register reading
other than `"Unlimited"` as `curtailed`. With a configured export cap, the
register legitimately reads a watt limit (e.g. `"Limited to 10000W"` for a
10 kW cap) in normal, healthy steady-state operation — so the sensor
reported `curtailed` continuously, regardless of whether real curtailment
was happening. This matches the reporter's exact scenario: a 10 kW grid
export limit, `Hardware Writes = Allowed`, `Plan Strategy = milp`, and the
inverter register unchanged at 10 kW while the sensor claims `curtailed`.

(The applier's own behavior of not curtailing below the DNO cap for
low-but-positive prices is correct per #767 and is unchanged by this PR —
only the diagnostic sensor's interpretation of the register was wrong.)

## Fix

`_is_directly_limited()` now resolves the expected routine baseline from
`cfg.max_grid_export_power_kw`:

- No cap configured → baseline is `"Unlimited"`; any limit is curtailment
  (unchanged legacy behaviour).
- A cap is configured → baseline is that cap in watts. Only a reported watt
  limit **below** the configured cap (with a small tolerance for Modbus
  read-back rounding) is reported as `curtailed` — e.g. the genuine
  negative-export-price 100 W block. A limit matching the cap is `normal`.
- A percentage-based limit or an unparseable limit string while a watt cap
  is configured is still reported as `curtailed` (unexpected/non-routine
  state, surfaced for visibility).

The derived (fallback, SoC + price heuristic) detection method is
unchanged.

## Files changed

- `custom_components/hsem/custom_sensors/pv_curtailment_sensor.py` — fixed
  `_is_directly_limited()`, now takes `cfg: SensorConfig | None` and reuses
  the existing `_parse_power_control_pct` / `_is_watt_limit` helpers from
  `applier_state_readers.py`.
- `tests/sensors/test_pv_curtailment_sensor.py` — new test file (none
  existed before): unit tests for `_is_directly_limited` and
  `_is_derived_curtailment`, plus end-to-end `state` tests including a
  direct regression reproduction of the issue #924 scenario.

## Tests

- `./scripts/quality.sh lint` — pass
- `./scripts/quality.sh typing` — pass (0 errors)
- `./scripts/quality.sh quality` — pass (0 pyright errors; vulture output
  unchanged baseline)
- `./scripts/quality.sh test` — pass (3122 passed, including 19 new tests)
- `python3 scripts/validate_translations.py` — 0 errors (no user-facing
  strings changed)

## Known limitations

- This PR only fixes the diagnostic sensor's misreporting. It does not
  change the applier's export behaviour — PV export below the configured
  `export_electricity_min_price` but at a non-negative price is still
  allowed up to the DNO cap by design (#767). If the reporter's actual goal
  is to have HSEM physically curtail PV production for low-but-positive
  export prices (not just report on it), that would be a separate,
  intentional policy change and should be tracked as its own issue.

Fixes #924

### DIFF
--- a/custom_components/hsem/custom_sensors/pv_curtailment_sensor.py
+++ b/custom_components/hsem/custom_sensors/pv_curtailment_sensor.py
@@ -8,7 +8,12 @@ The sensor uses two complementary detection methods:
 
 1. **Direct**: Reads ``live.huawei_inverter_active_power_control``.
    If the inverter reports a limit (e.g. ``"Limited to 80%"`` or
-   ``"Limited to 100W"``) instead of ``"Unlimited"``, PV is being curtailed.
+   ``"Limited to 100W"``) that sits *below* the routine, always-applied
+   ``max_grid_export_power_kw`` DNO/grid cap (or below ``"Unlimited"`` when no
+   cap is configured), PV is being actively curtailed. A limit that merely
+   matches the configured cap is normal steady-state operation, not
+   curtailment (issue #924) — the applier writes that cap for every
+   non-negative export price regardless of price-based curtailment.
 
 2. **Derived**: When the battery SoC is high (≥ 95 %) AND the export price
    is below the minimum threshold, curtailment is likely even if the direct
@@ -30,7 +35,12 @@ from custom_components.hsem.coordinator import (
     CoordinatorData,
     HSEMDataUpdateCoordinator,
 )
+from custom_components.hsem.custom_sensors.applier_state_readers import (
+    _is_watt_limit,
+    _parse_power_control_pct,
+)
 from custom_components.hsem.entity import HSEMCoordinatorEntity, HSEMEntity
+from custom_components.hsem.models.sensor_config import SensorConfig
 from custom_components.hsem.utils.sensornames.diagnostics import (
     get_pv_curtailment_sensor_entity_id,
     get_pv_curtailment_sensor_unique_id,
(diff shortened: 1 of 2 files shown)


woopstar:
@tonnr, could you please validate that the PV Curtailment sensor now reports normal when the inverter is merely at your configured 10 kW export cap, and curtailed only when HSEM actually throttles below that cap?

PR #929, opened 2026-09-06
fix(sensors): stop PV curtailment false positive at low positive export price

### LINKED ISSUE #925: MILP plans PV export below configured minimum export price while PV Curtailment reports curtailed
Hi,

I’m seeing inconsistent behavior in the current beta.

My configured minimum export price is:

export_electricity_min_price = 0.02

Current export price:

0.004

However, the MILP planner still plans:

PV export planned: 2.365 kWh
Grid export planned: 3.921 kWh

At the same time:

PV Curtailment = curtailed
Hardware Writes = allowed
Applier status = pending
Total writes = 0
Apply status = skipped
Failed entities = []
Unverified entities = []

So HSEM appears to detect that PV should be curtailed, but the planner still schedules export below the configured minimum price, and no inverter write is attempted.

This looks more like a planner/curtailment inconsistency than a Huawei write failure.

Could you please check whether export_electricity_min_price is supposed to be a hard floor for PV export in the MILP planner, and why PV Curtailment = curtailed does not result in any applier write?

woopstar:
## Summary

Fixes #925 — "MILP plans PV export below configured minimum export price while PV Curtailment reports curtailed".

- The reporter's scenario (`export_electricity_min_price = 0.02`, live export price `0.004`, `Applier status = pending`, `Total writes = 0`) is actually **correct behaviour**: since issue #767, `export_electricity_min_price`/`batteries_export_min_price` only gate *intentional battery-to-grid export*. PV surplus export is never blocked at a non-negative price — only a **negative** export price makes the applier physically block the whole grid connection point. The MILP/applier code already implements this correctly (`planner/milp/_price_sanitise.py`, `custom_sensors/applier_power_control.py`).
- The real bug: `custom_sensors/pv_curtailment_sensor.py`'s derived-detection heuristic still used the **pre-#767** assumption — it flagged `"curtailed"` whenever the export price was below a fixed `0.01` threshold, the battery SoC was ≥ 95 %, PV was producing, and the inverter's active-power-control register was unavailable. That's exactly the reporter's scenario, and it produced a false `"curtailed"` reading even though the planner/applier were correctly continuing to export the PV surplus.
- Also corrected `hsem_export_electricity_min_price` translation copy (en/da/de/es) and two `docs/troubleshooting-guide.md` sections, which still described the old "blocks all export" behaviour and were the likely source of the reporter's expectation.

## Changes

- `custom_components/hsem/custom_sensors/pv_curtailment_sensor.py`: derived heuristic now keys off a **negative** export price (matching `applier_power_control.py`'s real gating boundary) instead of a fixed low-price threshold. Updated module/function docstrings accordingly.
- `custom_components/hsem/translations/{en,da,de,es}.json`: rewrote the `hsem_export_electricity_min_price` description to state it only gates intentional battery-to-grid export and does not block PV surplus export.
- `docs/troubleshooting-guide.md`: sections 2c/7c corrected to describe the current (post-#767) behaviour.
- `tests/custom_sensors/test_pv_curtailment_sensor.py` (new): regression coverage for the exact issue-925 scenario, plus negative-price detection and direct-register-override cases.

## Test plan

- [x] `pytest tests/custom_sensors/test_pv_curtailment_sensor.py` — 10 passed
- [x] `./scripts/quality.sh test` — 3113 passed
- [x] `python3 scripts/validate_translations.py` — 0 errors (pre-existing unrelated warnings only)

### DIFF
--- a/custom_components/hsem/custom_sensors/pv_curtailment_sensor.py
+++ b/custom_components/hsem/custom_sensors/pv_curtailment_sensor.py
@@ -2,7 +2,13 @@
 
 Curtailment occurs when the inverter throttles solar production because there
 is no place for the energy to go — the battery is full, house consumption is
-low, and grid export is blocked (export price below the configured minimum).
+low, and grid export is physically blocked. Since issue #767, HSEM only
+physically blocks the whole grid connection point when the export price is
+**negative** (exporting would cost money); a positive-but-low export price
+below ``export_electricity_min_price``/``batteries_export_min_price`` only
+gates intentional *battery*-to-grid export, PV surplus keeps exporting
+normally. Do not conflate "export price below the configured minimum" with
+"grid export is blocked" — see ``applier_power_control.py``.
 
 The sensor uses two complementary detection methods:
 
@@ -11,8 +17,8 @@ The sensor uses two complementary detection methods:
    ``"Limited to 100W"``) instead of ``"Unlimited"``, PV is being curtailed.
 
 2. **Derived**: When the battery SoC is high (≥ 95 %) AND the export price
-   is below the minimum threshold, curtailment is likely even if the direct
-   register has not yet updated.
+   is negative (the only case where HSEM itself blocks grid export),
+   curtailment is likely even if the direct register has not yet updated.
 
 The sensor state is either ``"curtailed"`` or ``"normal"``.
 """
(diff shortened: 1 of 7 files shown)


woopstar:
@tonnr, could you please confirm that, at a low but positive export price, the PV Curtailment sensor now reports normal while PV surplus continues to export as intended by the existing policy?

PR #932, opened 2026-09-06
fix(diagnostics): log which slot/field trips invalid_future_values

### LINKED ISSUE #931: fix(diagnostics): log which slot/field trips the invalid_future_values safety hold
## Description

Reported as a follow-up comment on #925: a user saw `hourly_recommendations`
contain 192 slots (48h horizon) where only 1 slot had a `recommendation` and
191 slots were `recommendation=None`. `sensor.hsem_working_mode` →
`data_quality` showed `load_forecast_ready=False`,
`load_forecast_reason=invalid_future_values`, and the coordinator published a
`safety_hold` plan instead of a normal MILP plan.

This fail-closed behaviour is *working as designed* —
`assess_load_forecast()` in `custom_components/hsem/coordinator_helpers.py:402-460`
deliberately holds the battery rather than planning on a broken future-load
profile (see the `LoadForecastReadiness` docstring and
`apply_load_forecast_hold` in the same module). The problem is that the
`"invalid_future_values"` reason string (set at
`coordinator_helpers.py:434` and `:437`) gives operators — and us — no way to
tell **which** of the 192 slots or **which** of the 5 fields
(`avg_house_consumption_kwh`, `_1d_kwh`, `_3d_kwh`, `_7d_kwh`, `_14d_kwh`)
actually failed validation, or what the offending raw value/type was. Without
that, a genuine upstream bug in the ML consumption predictor
(`ml/populator.py`) or the legacy average-sensor path
(`custom_sensors/hourly_data_populator/consumption.py`) that occasionally
produces a NaN/negative/non-numeric value is completely invisible in the logs
— we only ever see the generic reason, never the value that tripped it.

## Possible Solution

Add diagnostic logging at the exact point each validation branch fails in
`assess_load_forecast()`, using the existing `async_log()` helper (already
imported in this module and safe to call from synchronous code — see
`utils/logger.py::log_planner`/`async_log`). Log:

- The failing slot's timestamp (`utc_key(rec.start).isoformat()`).
- The specific field name that failed (not just "some future value").
- For non-numeric/bool values: the raw value's `repr()` and `type(...).__name__`.
- For out-of-range values (non-finite or negative): the numeric value itself.

This is purely additive — it must not change the fail-closed return values or
any planner/safety semantics (do not touch the validation conditions
themselves, only add logging immediately before each existing
`return LoadForecastReadiness(False, "invalid_future_values", None)`).

## Acceptance Criteria

- [ ] When a future slot fails validation, the HSEM log records the exact
      slot timestamp and the specific field name that failed.
- [ ] Non-numeric/bool failures log the value's type; out-of-range failures
      log the numeric value.
- [ ] `assess_load_forecast()`'s return values and fail-closed behaviour are
      byte-for-byte unchanged — this is a logging-only change.
- [ ] Tests added covering both the non-numeric and out-of-range branches,
      asserting the log call contains the field name and slot timestamp.
- [ ] `./scripts/quality.sh all` passes.

## Suggested Prompt

```
Read custom_components/hsem/coordinator_helpers.py's assess_load_forecast()
function (around line 402-460) and its LoadForecastReadiness dataclass. It
currently returns a bare "invalid_future_values" reason string with no detail
about which slot/field caused the failure, making real-world reports (see
issue #925 comment thread) impossible to root-cause from logs alone.

Add a warning-level async_log() call immediately before each of the two
`return LoadForecastReadiness(False, "invalid_future_values", None)`
statements, naming the specific field (avg_house_consumption_kwh,
_1d_kwh, _3d_kwh, _7d_kwh, or _14d_kwh) and the slot's
utc_key(rec.start).isoformat() timestamp. For the non-numeric/bool branch,
include repr(raw_value) and type(raw_value).__name__; for the
non-finite/negative branch, include the numeric value. Do not alter the
validation logic or return values — this is diagnostics-only.

Add tests in tests/test_load_forecast_readiness.py that patch
custom_components.hsem.coordinator_helpers.async_log and assert the logged
message contains the expected field name and slot timestamp for both failure
branches.

Run the hsem-pre-flight skill first (branch off main), then
./scripts/quality.sh lint/typing/quality/test before opening the PR.
```

woopstar:
## Summary

Fixes #931 — follow-up from the second comment on #925 (191/192 `hourly_recommendations` slots had `recommendation=None` because `load_forecast_reason=invalid_future_values` engaged a `safety_hold`).

That fail-closed behaviour is correct by design — `assess_load_forecast()` deliberately holds the battery rather than plan on a broken future-load profile. The gap was purely diagnostic: the generic `"invalid_future_values"` reason gave no way to tell which of the 192 slots, or which of the 5 consumption fields, actually failed validation, so a genuine upstream ML/legacy-avg data bug would be invisible in the logs.

## Changes

- `custom_components/hsem/coordinator_helpers.py::assess_load_forecast()`: added a warning-level `async_log()` call at each existing fail-closed return, naming the slot's UTC timestamp and the specific field (`avg_house_consumption_kwh`/`_1d_kwh`/`_3d_kwh`/`_7d_kwh`/`_14d_kwh`). Non-numeric/bool failures log the raw value and its type; non-finite/negative failures log the numeric value. No change to the validation logic or return values — logging only.
- `tests/test_load_forecast_readiness.py`: two new tests asserting the log call fires with the correct field name, slot timestamp, and type/value detail for both failure branches.

## Test plan

- [x] `pytest tests/test_load_forecast_readiness.py` — 27 passed
- [x] `./scripts/quality.sh test` — 3115 passed

### DIFF
--- a/custom_components/hsem/coordinator_helpers.py
+++ b/custom_components/hsem/coordinator_helpers.py
@@ -421,19 +421,36 @@ def assess_load_forecast(
     signature_slots: list[_LoadSlotSignature] = []
     weighted_profile_is_zero = True
     for rec in future_slots:
-        raw_values = (
-            rec.avg_house_consumption_kwh,
-            rec.avg_house_consumption_1d_kwh,
-            rec.avg_house_consumption_3d_kwh,
-            rec.avg_house_consumption_7d_kwh,
-            rec.avg_house_consumption_14d_kwh,
+        named_raw_values = (
+            ("avg_house_consumption_kwh", rec.avg_house_consumption_kwh),
+            ("avg_house_consumption_1d_kwh", rec.avg_house_consumption_1d_kwh),
+            ("avg_house_consumption_3d_kwh", rec.avg_house_consumption_3d_kwh),
+            ("avg_house_consumption_7d_kwh", rec.avg_house_consumption_7d_kwh),
+            ("avg_house_consumption_14d_kwh", rec.avg_house_consumption_14d_kwh),
         )
         canonical_values: list[float] = []
-        for raw_value in raw_values:
+        for field_name, raw_value in named_raw_values:
             if isinstance(raw_value, bool) or not isinstance(raw_value, (int, float)):
+                async_log(
+                    "warning",
+                    "[load] Forecast slot %s has a non-numeric %s (%r, type=%s); "
+                    "holding as invalid_future_values.",
+                    utc_key(rec.start).isoformat(),
+                    field_name,
+                    raw_value,
+                    type(raw_value).__name__,
+                )
                 return LoadForecastReadiness(False, "invalid_future_values", None)
             number = float(raw_value)
             if not math.isfinite(number) or number < 0.0:
+                async_log(
+                    "warning",
+                    "[load] Forecast slot %s has an out-of-range %s (%s); "
+                    "holding as invalid_future_values.",
+                    utc_key(rec.start).isoformat(),
+                    field_name,
+                    number,
+                )
                 return LoadForecastReadiness(False, "invalid_future_values", None)
             canonical_values.append(round(number, 5))
 
--- a/tests/test_load_forecast_readiness.py
+++ b/tests/test_load_forecast_readiness.py
@@ -92,6 +92,52 @@ class TestLoadForecastAssessment:
         )
         assert result.ready is False
         assert result.reason == "invalid_future_values"
+
+    def test_invalid_value_is_logged_with_slot_and_field_name(self) -> None:
+        # Issue #925 follow-up: the generic "invalid_future_values" reason
+        # alone gives operators nothing to act on. Diagnostic logging must
+        # name the offending slot and field so a bad ML/legacy-avg output
+        # can actually be traced back to its source.
+        now = datetime(2026, 8, 21, 12, 5, tzinfo=UTC)
+        profile = _profile(now, 0.2)
+        profile[1].avg_house_consumption_7d_kwh = float("nan")
+
+        with patch("custom_components.hsem.coordinator_helpers.async_log") as mock_log:
+            result = assess_load_forecast(
+                profile,
+                now,
+                population_succeeded=True,
+                live_house_demand_w=0.0,
+            )
+
+        assert result.reason == "invalid_future_values"
+        mock_log.assert_called_once()
+        level, msg, *args = mock_log.call_args.args
+        assert level == "warning"
+        formatted = msg % tuple(args)
+        assert "avg_house_consumption_7d_kwh" in formatted
+        assert profile[1].start.isoformat() in formatted
+
+    def test_non_numeric_value_is_logged_with_type(self) -> None:
+        now = datetime(2026, 8, 21, 12, 5, tzinfo=UTC)
+        profile = _profile(now, 0.2)
+        profile[1].avg_house_consumption_1d_kwh = None  # type: ignore[assignment]
+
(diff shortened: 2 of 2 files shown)


woopstar:
@tonnr, could you please confirm on the next safety hold that the new diagnostics identify the failing forecast slot and field, so we can pinpoint the invalid future value?

tonnr:
Yes. The system is currently in safety_hold, and I was able to narrow the invalid forecast down to the 1-day house-consumption field.

In the 48h hourly_recommendations horizon, I found:

avg_house_consumption_1d_kwh = -0.205

for the 14:00–15:00 slots on both 2026-09-06 and 2026-09-07.

The other consumption averages for the same slots remain positive:

avg_house_consumption_kwh = 0.331
avg_house_consumption_3d_kwh = 0.535
avg_house_consumption_7d_kwh = 0.362
avg_house_consumption_14d_kwh = 0.297

At the same time HSEM reports:

load_forecast_ready = False
load_forecast_reason = invalid_future_values
horizon_has_tomorrow = False

So the likely failing field is avg_house_consumption_1d_kwh for those slots.

I also have a likely explanation for how the negative value was introduced. While testing the new HSEM configuration, I briefly changed the consumption/EV accounting settings incorrectly and enabled net-consumption based accounting in a setup where EV consumption was also being handled separately. This appears to have produced the negative historical consumption value for the 14:00–15:00 period.

After correcting the configuration, the source consumption sensor looks normal again, but the negative 1-day historical value remained in HSEM’s forecast inputs and appears to be what triggered invalid_future_values / safety_hold.

So the original invalid value may have been caused by my temporary misconfiguration, but it would still be very useful if HSEM could:

* identify the exact failing slot and field in the diagnostics;
* recover cleanly once the source configuration has been corrected.

One more important observation: while HSEM is in this state, the Huawei Maximum discharging power setting is repeatedly toggling between 5000 W and 0 W, roughly every minute.

This does not look healthy and it appears to have a real effect on the system. During the night, the battery SoC decreased, but part of the house load was still imported from the grid. If the discharge limit is being set to 0 W periodically, that would explain why the battery temporarily stops supporting the house and grid import occurs.

Could you please check whether safety_hold or the fallback/applier logic is intentionally writing the maximum discharge power in this way? I would not expect it to oscillate continuously between 0 W and 5000 W.

I can also validate the new safety-hold diagnostics against this exact case once the updated version is available.

woopstar:
Thanks for the detailed data, @tonnr — that's exactly what the new diagnostics were meant to surface, and it pointed straight to a root cause. Filed two follow-up issues, since they're separate bugs with separate fixes:

- #938 — why `safety_hold` didn't clear after you fixed the source config: the "1d" rolling-average sensor (`HSEMAvgSensor`) never validates that a stored day's reading is non-negative before persisting it, so the bad `-0.205` sample from the misconfiguration became the sole "1d" measurement and stuck around until that specific hour block completed again on a later day — the fix rejects negative/non-finite readings before they enter the rolling window.
- #939 — the max-discharge-power oscillation: `async_apply_battery_settings()` in `applier.py` unconditionally writes the rated max discharge power first, then separately re-caps it to 0W later in the same call whenever a hold condition (like `safety_hold`) is active. That's a guaranteed write-then-undo every apply cycle, which matches the ~1-minute 5000W/0W toggling you saw and is the likely explanation for the overnight grid import alongside SoC decrease. Filed as `priority:p0` since it's a live hardware-control bug, not just diagnostics.

I'll ping this thread once both land so you can validate against your setup — appreciate you narrowing it down this precisely.

Issue #914, opened 2026-09-03
fix(planner): self_consumption_with_reserve reserve stops at first PV-surplus slot, ignoring plan's future discharge


woopstar:
## Description

When `batteries_wait_mode_behavior` is `self_consumption_with_reserve`, the reserve
that protects battery energy during a `BatteriesWaitMode` slot is computed by
`calculate_required_battery_until_solar()` in
`custom_components/hsem/planner/discharge_scheduler.py:155-203`. That function scans
forward from `now` and accumulates positive net-consumption **only until the first
slot with negative net-consumption (any forecast PV surplus) is found**, then breaks
(line 182-183):

```python
if slot.estimated_net_consumption_kwh < 0:
    break
if slot.estimated_net_consumption_kwh > 0:
    required += slot.estimated_net_consumption_kwh
```

This value is assigned to `planner_output.required_capacity_kwh` in
`custom_components/hsem/planner/engine_core.py:189`, propagated to
`coordinator_planner_phase.py:196` / `:225` as `self._current_required_battery`, and
finally consumed by the applier in
`custom_components/hsem/custom_sensors/applier.py:394-402` and `:468-486` as
`current_required_battery_kwh` — the reserve threshold used to decide whether the
battery may discharge during a Wait slot in self-consumption mode:

```python
surplus = live.battery_current_capacity_kwh - current_required_battery_kwh
if surplus > 1e-9:
    working_mode = WorkingModes.MaximizeSelfConsumption.value
```

The function name and docstring ("Estimate battery capacity needed until the first
solar surplus slot") describe an *excess-export* helper, not a wait-mode reserve
calculation — it is also used for that purpose via `apply_excess_export()` further
down `discharge_scheduler.py`. Using it to gate wait-mode discharge is a
mismatch: it stops as soon as it sees the *first* forecast PV-surplus slot, no matter
how small or short-lived that surplus is, and regardless of whether the MILP's
selected plan actually expects to be able to recharge from it. It does not account
for the selected MILP plan's future planned discharge, so it does not guarantee the
plan's SoC trajectory remains followable.

Reported by @AndersN76 in
https://github.com/woopstar/hsem/issues/732#issuecomment-5527513303, reproduced on
HSEM 6.2.4 (this is a follow-up/regression on #742 / PR #743, which introduced
`self_consumption_with_reserve` specifically to preserve battery capacity for future
expensive periods). Observed example: at ~12:30 the MILP recommendation was Wait with
[...]

PR #950, opened 2026-09-08
fix(planner): stop wait-mode reserve scan at next discharge, not just next charge

### LINKED ISSUE #942: Low battery discharge cap causes unnecessary grid import during load spikes
Issue body:

While validating v6.3.0-beta1 fixes around #934 and #941, I found a separate battery-control issue.

HSEM is setting Huawei Maximum discharging power to low forecast-based values instead of leaving the battery available at its configured/rated maximum (5000 W).

Observed:

19:15 → 264 W
19:30 → 376 W
19:45 → 380 W
20:00 → 376 W

I then switched on the stove while the discharge cap was ~380 W.

Result:

battery discharge did not increase above the cap;
the extra household load was supplied from the grid;
the discharge cap did not react to the real load increase.

So the forecast-based discharge cap can cause unnecessary grid import even when the battery has enough SoC and can physically discharge up to 5000 W.

Expected behavior: normal house-load support should remain available up to the configured/rated battery discharge limit, while battery-to-grid export is controlled separately by export-price / curtailment logic.

Related:

#941 — max-discharge-power write/cap logic
#934 — export-fee based curtailment
#930 / #935 — below-minimum-price PV curtailment

This looks like a separate hardware-control correctness issue from the 5000W/0W oscillation fixed in #941.

woopstar:
## Summary

- Follow-up to #949: after that PR shipped the SoC-floor discharge-cap gate,
  `tonnr` reported (https://github.com/woopstar/hsem/issues/942#issuecomment-5584381684)
  that the gate was still writing `0 W` for the entire `14:30`–`19:30` Wait
  span, even with the battery at 100 %.
- Root cause: `calculate_required_battery_for_plan()` (`discharge_scheduler.py`)
  only stopped its forward SoC-trajectory scan at the plan's next **charge**
  event. Since Wait-mode slots never discharge in the simulation
  (`soc_simulation.py` forces `discharge = 0.0` while
  `recommendation == BatteriesWaitMode`), the scan instead accumulated
  through *every* future discharge slot up to that charge — in this report,
  the plan's whole overnight total — and applied that sum as an immediate
  reserve floor hours before any of it was actually needed. That collapsed
  `battery_current_capacity_kwh - wait_mode_reserve_kwh` to ~0 for the whole
  afternoon, silently reopening #942 even though the #949 gate itself was
  working exactly as designed.
- Fix: the scan now also breaks on the plan's next **discharge**
  (`batteries_discharged_kwh > 1e-9`), so the reserve protects only the
  plan's very next committed battery action. Anything further out is
  re-derived fresh by the next replan (interval tick, event-triggered, or
  the 10-second live-power monitor) once it actually becomes relevant —
  matching the direction confirmed in the issue thread.
- `docs/planner-spec.md` §"Wait-mode self-consumption reserve (issue #914)"
  and `.github/memories.md` updated to document the new stop condition.

## Test plan

- [x] Added `TestReserveStopsAtFirstScheduledAction` to
      `tests/planner/test_wait_mode_reserve_from_plan.py`, reproducing the
      reported scenario (long flat Wait run → small first discharge →
      further overnight discharge slots) and asserting the reserve now
      equals only the first discharge's dip, not the cumulative overnight
      total. Also covers the unchanged "charge comes first" case.
- [x] `./scripts/quality.sh test` (3306 passed)
- [x] `./scripts/quality.sh translations` (no user-facing strings changed)

Fixes #942

### DIFF
--- a/.github/memories.md
+++ b/.github/memories.md
@@ -1308,6 +1308,20 @@ instead of keeping the battery strictly idle.
   means no reliable reserve could be derived; the applier then forces strict
   TOU wait instead of enabling self-consumption. See `docs/planner-spec.md`
   §"Wait-mode self-consumption reserve (issue #914)".
+- **Scan stops at the next discharge too, not just the next charge (issue
+  #942 follow-up, fixed 2026-09-08):** the scan originally broke only on a
+  genuine planned _charge_, so it accumulated through every future discharge
+  slot up to that charge — often the plan's whole overnight total — and
+  applied that sum as an immediate floor hours before it was needed. Since
+  Wait-mode slots never discharge in the simulation (`soc_simulation.py`
+  forces `discharge = 0.0` while `recommendation == BatteriesWaitMode`),
+  this silently starved the #942 SoC-floor gate of any surplus for the
+  entire Wait span (reserve ≈ current capacity), even though the fix had
+  just landed. The loop now also breaks on `batteries_discharged_kwh > 1e-9`:
+  the reserve protects only the plan's very next committed action (charge or
+  discharge), trusting the next replan (interval tick, event-triggered, or
+  the 10-second live-power monitor) to re-derive the reserve fresh from the
+  then-current capacity before any later slot arrives.
 
 Files involved: `flows/batteries_wait_mode.py`, `config_flow.py`,
 `options_flow.py`, `translations/en.json`, `const.py`,
--- a/custom_components/hsem/planner/discharge_scheduler.py
+++ b/custom_components/hsem/planner/discharge_scheduler.py
@@ -217,10 +217,20 @@ def calculate_required_battery_for_plan(
     :func:`~custom_components.hsem.planner.soc_simulation.simulate_soc` for
     the winning candidate) and returns how far that trajectory dips below
     ``current_capacity`` before the plan's next slot with an actual, solved
-    battery charge (grid or solar) — not just a forecast surplus. A small or
-    short-lived forecast surplus that the plan does not actually charge from
-    does not end the scan early, so the reserve still protects a later
-    planned discharge.
+    battery **action** — charge (grid or solar) or discharge — not just a
+    forecast surplus. A small or short-lived forecast surplus that the plan
+    does not actually charge from does not end the scan early, so the
+    reserve still protects the plan's very next committed discharge.
+
+    The scan stops at the first committed action rather than accumulating
+    through every future discharge slot up to the next charge (issue #942
+    follow-up): reactive replanning (interval tick, event-triggered, or the
+    10-second live-power monitor) re-derives this same reserve from the
+    then-current capacity before any later slot arrives, so protecting
+    slots beyond the very next one here would only lock up capacity for
+    house-load self-consumption hours before it's actually needed, without
+    adding real protection — the next replan supersedes this value long
+    before that later slot starts.
 
     Slots are sorted by start time before scanning, so calling code does not
     need to guarantee chronological order.
@@ -245,7 +255,7 @@ def calculate_required_battery_for_plan(
     min_capacity = current_capacity
     for slot in future_slots:
         min_capacity = min(min_capacity, slot.estimated_battery_capacity_kwh)
-        if slot.batteries_charged_kwh > 1e-9:
+        if slot.batteries_charged_kwh > 1e-9 or slot.batteries_discharged_kwh > 1e-9:
             break
 
     result = round(max(current_capacity - min_capacity, 0.0), 3)
--- a/docs/planner-spec.md
+++ b/docs/planner-spec.md
@@ -2136,19 +2136,37 @@ from the **selected** plan's own already-simulated SoC trajectory instead:
 ```text
 for each future slot (chronological order, end > now):
     min_capacity = min(min_capacity, slot.estimated_battery_capacity_kwh)
-    if slot.batteries_charged_kwh > 0:  # a genuine solved charge, not a
-        break                           # forecast surplus signal
+    if slot.batteries_charged_kwh > 0 or slot.batteries_discharged_kwh > 0:
+        break  # a genuine solved battery action, not a forecast signal
 reserve = max(current_capacity - min_capacity, 0)
 ```
 
 This protects the battery down to the deepest point the **winning
-candidate's** SoC simulation actually dips to before its own next solved
-charge (grid or solar) — a small forecast surplus slot that the plan does not
-actually charge from no longer truncates the reserve early. When the plan
-has no future charge anywhere in the horizon, the scan runs to the horizon
-end and the reserve naturally covers (up to) the full current capacity,
-which forces strict Wait behaviour via the normal `surplus <= 0` path — no
-special-cased fallback is needed for that case.
+candidate's** SoC simulation dips to by the end of its own next solved
+battery action (charge — grid or solar — **or** discharge) — a small
+forecast surplus slot that the plan does not actually charge from no longer
+truncates the reserve early. When the plan has no future charge or discharge
+anywhere in the horizon, the scan runs to the horizon end and the reserve
+naturally covers (up to) the full current capacity, which forces strict Wait
+behaviour via the normal `surplus <= 0` path — no special-cased fallback is
+needed for that case.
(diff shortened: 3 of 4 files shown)


tonnr:
I think the expected behavior should be simpler and stricter:

If the user has configured Huawei Maximum discharging power = 5000 W, HSEM should not overwrite that value with lower forecast-derived caps during normal operation.

That setting represents the maximum discharge power the user is willing to allow. The inverter can still use less power depending on actual house demand, but HSEM should leave the physical ceiling at the configured value.

Otherwise HSEM can create unnecessary grid import: if the planner caps discharge to 0 W or a few hundred watts and the house load suddenly increases, the inverter cannot respond from the battery and the missing power is bought from the grid at the current import price.

My preference is:

keep Maximum discharging power at the user-configured value (5000 W);
let normal self-consumption use the battery up to that limit;
control battery-to-grid export separately with export-price / curtailment logic;
protect planned reserve using an SoC/energy floor, not by lowering the user's physical discharge-power ceiling.

In short: HSEM may decide when discharge is allowed, but it should not silently redefine the user's configured maximum discharge capability in a way that causes avoidable grid import.

woopstar:
You're right, and I found exactly why: this goes deeper than #949/#950.

**Root cause:** `_primary_battery_hold(rec)` returns `True` for essentially every genuine `BatteriesWaitMode` slot — "Wait" means the plan scheduled neither charge nor discharge, which is the exact same near-zero-energy condition the hold check tests for (`soc_simulation.py` forces `discharge = 0.0` whenever `recommendation == BatteriesWaitMode`). In `applier.py`, the cap computation checks `primary_battery_hold` *first* and unconditionally sets the discharge cap to `0 W` when it's true — **before** the `self_consumption_with_reserve` reserve-floor logic ever gets a chance to run (that logic is gated on `not primary_battery_hold`). So for real Wait slots, the reserve-floor gate from #949/#950 was essentially dead code: the hold check always won first.

Reproduced directly against current `main`: battery at 100% (5.0 kWh), reserve = 1.0 kWh (4 kWh of surplus available), genuine held Wait slot (`batteries_charged_kwh=0`, `batteries_discharged_kwh=0`) → the applier still writes `Maximum discharging power = 0 W`, matching what you're seeing.

**Fix direction**, matching what you described: the `self_consumption_with_reserve` reserve-floor decision needs to run *instead of* the hold-to-0 default (not after it) whenever the user has opted into that mode, for genuine Wait slots with no active/planned EV and no authoritative held export. Concretely: fold the reserve-floor check into the single cap decision ahead of the hold check, so `primary_battery_hold` only forces `0 W` when `self_consumption_with_reserve` isn't active (or the reserve isn't derivable) — matching your ask of "protect the reserve with an SoC floor, don't lower the user's configured ceiling for normal self-consumption."

Working on the fix now; will link the PR here.

woopstar:
Opened #955 with the fix.

To directly address your ask: the discharge-power register is no longer lowered as a *rate*, and the `self_consumption_with_reserve` reserve-floor decision now runs regardless of whether the slot is "held" — which real Wait slots always are, and which is exactly why #949/#950 didn't take effect for you. Concretely, once #955 is merged:

- The user-configured/rated maximum is what gets written whenever the battery holds any surplus above the reserve — normal self-consumption is never throttled to a lower wattage.
- Once capacity reaches the reserve floor, the register is written to `0 W` (an explicit stop, not a lowered ceiling) — the reserve is protected by that floor, not by redefining your configured maximum.
- Battery-to-grid export stays governed separately by export-price/curtailment logic, untouched by any of this.

Reproduced your exact scenario as a regression test (100% SoC, held Wait slot, reserve well below capacity) before and after the fix — see #955 for details.

tonnr:
I’m still seeing the same practical issue on my setup after the latest change:

Huawei Maximum discharging power remains at 0 W
the house is importing from the grid
the battery still has significant SoC available

I also want to highlight the existing user setting:

Battery Wait Mode Behaviour = Self-consumption with reserve

My understanding of this option is that when HSEM selects Wait mode, the house should still be allowed to use battery energy for normal self-consumption, while only the planner reserve is protected.

In other words, this setting should mean:

Wait mode does not mean “do not use the battery at all”
normal house load should still be supplied from the battery
only the reserved energy floor should be protected
HSEM should avoid unnecessary grid import while battery energy above the reserve is available

Right now, despite having Self-consumption with reserve enabled, HSEM still sets Maximum discharging power = 0 W, so the house is forced to import from the grid.

That seems inconsistent with the purpose of the setting itself. I think this option should be the authoritative user preference for Wait-mode behavior: if Self-consumption with reserve is selected, the battery should continue covering house load up to the user-configured discharge limit, and only stop once the actual reserve floor is reached.

If Strict wait is selected, then 0 W would make sense. But with Self-consumption with reserve, I would expect house consumption to continue being served from the battery.

woopstar:
This matches #955 exactly (0 W cap, house importing despite available SoC) — but that fix merged at 09:59 UTC today, and `v6.3.0-beta3` was cut at 08:31 UTC, *before* the merge. So beta3 (and anything earlier) still has the precedence bug; what you're testing doesn't include the fix yet.

`v6.3.0-beta3` includes #950 (the reserve-scan fix) but not #955 (the hold-precedence fix). You'll need the next pre-release build (once one is cut including commit `d1d4e25`) to see #955's effect — could you re-test once that's available?

tonnr:
That explains it — I was testing v6.3.0-beta3, so #955 was not included yet.

I’ll re-test as soon as the next pre-release build is available with commit d1d4e25.

I’ll specifically verify that with Self-consumption with reserve selected:

Huawei Maximum discharging power stays at my configured 5000 W while battery capacity is above the reserve floor;
normal house load is served from the battery instead of the grid;
discharge only drops to 0 W once the reserve floor is actually reached.

woopstar:
`v6.3.0-beta4` is out for testing and includes the fix from #955 (commit `d1d4e25`).

Release: https://github.com/woopstar/hsem/releases/tag/v6.3.0-beta4

Could you re-test with this build? Specifically: a Wait-mode slot with `Self-consumption with reserve` enabled, battery SoC well above the reserve — the discharge cap should now stay at the configured/rated maximum and serve normal house load from the battery, only dropping to 0 W once the reserve floor is actually reached.

tonnr:
I re-tested on v6.3.0-beta4 with #955 included, but I am still seeing the same practical behavior.

Current state:

Working mode = batteries_wait_mode
Battery current capacity = 6.4 kWh
Battery usable capacity = 9.5 kWh
Battery SoC = 69%
Battery discharge planned = 0.0 kWh
Grid import planned = 0.192 kWh

Huawei Maximum discharging power is still being set to 0 W.

So with Self-consumption with reserve enabled, the battery is still prevented from serving normal house load while HSEM plans grid import instead.

This suggests that the remaining issue is not the discharge-rate cap itself anymore, but the effective reserve floor: it appears to equal the full current battery capacity in this Wait slot, leaving no “surplus above reserve” for self-consumption.

Could you please check how wait_mode_reserve_kwh is being derived in this case? From the user perspective, Self-consumption with reserve should allow normal house load to use the battery until the actual planned reserve floor is reached, rather than treating the entire current battery capacity as reserved.

woopstar:
Found it — this is a different, deeper layer of the same reserve mechanism. Opened #956 with the analysis and #958 with the fix.

Root cause: `calculate_required_battery_for_plan()` protects the *full* energy needed for the plan's next committed action, no matter how far away that action is. Reproduced your numbers' shape directly: battery at 6.4 kWh with a discharge scheduled 5 hours away needing 5.9 kWh computes a reserve of `5.9 kWh` **right now** — leaving only `0.5 kWh` of surplus for hours before the battery is actually needed. That's exactly the pattern you're seeing (69% SoC, 0 kWh discharge planned for the current slot, but the reserve is still consuming almost the whole capacity because of what's coming later).

Fix: the reserve now time-decays — it's only protected in full once the next action is imminent (within 2 hours), ramping down to zero protection the farther away it is, trusting the next replan to re-tighten it as the action approaches. So a discharge hours away no longer blocks self-consumption in the meantime; it only starts constraining the battery once it's actually close.

Will post here again once a beta build including #958 is out for you to re-test.

Issue #988, opened 2026-09-13
fix(planner): unavailable EV SoC falls back to 0 %, causing peak-price charging past target SoC


woopstar:
## Description

An EV that was already **above** its target SoC charged to 100 % straight through the
evening price peak. Reported behaviour on 2026-09-12: charging started ≈ 19:45 and ran to
100 %, with force charge **disabled**, smart charging **enabled**, EV SoC ≈ **83 %** and EV
target SoC = **80 %**. HSEM should never have requested charging.

### What the logs show

The last cycle before the incident is unambiguous — HSEM correctly declined to charge:

```
19:19:33  [core] _build_and_inject_for_ev  label=primary  connected=True  smart=True
          soc=83.0%  target=80.0%  cap=86.50  pwr=11.00  eff=92.0%  min_pwr=1380W
19:19:33  [ev_planner] build_ev_charging_plan START  soc=83.0%  target=80.0%
          needed=0.000kWh  deadline=2026-09-13T07:00:00+02:00
19:19:33  [ev_planner] build_ev_charging_plan DONE  state=fully_charged
          energy_needed=0 (already at or above target SoC)
19:19:35  [soc_sim] 12 19:45→20:00 ... ev_inj=0.000  ev_acc=0.000
19:19:35  Current hourly recommendation: state=batteries_discharge_mode
          ev_charger_calculated_power=0W  ev_total_planned_load_kwh=0.000
```

`ev_charger_calculated_power=0W` in **107/107** logged cycles on 09-12 and **132/132** on
09-13. The planned EV load for the 19:45 slot was `0.000`. HSEM also ranked
`12 19:15 … 12 20:45` as its *battery discharge* window — i.e. HSEM itself identified the
charging window as the day's most expensive slots.

Then, at **19:19:58**, Home Assistant restarted (`homeassistant.loader` integration-scan
banner at the head of `home-assistant.log.1`). Charging began ≈ 25 minutes later.

**Note on evidence:** `hsem.log` rotation dropped the 19:20 → 02:00 window, so the decisive
cycles are not in the supplied logs (`hsem.log.2` in the 09-13 capture starts at 02:00:03).
The root cause below is verified in code and matches the symptom and timing exactly, but it
is not directly witnessed in a log line. See *Acceptance Criteria* for the regression test
that would pin it.

### Root cause: unavailable EV SoC becomes 0 %

An unavailable EV SoC sensor is silently coerced to **0 %** — the most dangerous value it
could take.

[...]

Issue #775, opened 2026-08-20
[MILP] EV at target SoC still commanded to charge — house battery should take surplus PV first


woopstar:
## Summary

When the EV is at (or above) its target SoC with `ev_allow_charge_past_target_soc` enabled, the MILP can still command the EV charger to draw significant power from surplus PV **while the house battery is not full and is being charged in the same slot**. The house battery should take surplus PV first; the EV should only absorb what the battery cannot.

## Reproduction (v6.2.4, live diagnostics)

State at the current slot (15:30–15:45):

| field | value |
|---|---|
| `number.hsem_ev_target_soc` | 80 % |
| `sensor.audi_e_tron_state_of_charge` | 80 % |
| `ev_allow_charge_past_target_soc` | true |
| `ev_charger_status_state` | true (charging) |
| `ev_charger_power_state` | 3603 W (live) |
| `ev_charger_calculated_power` | **3815 W** |
| `ev_planned_load_kwh` / `ev_accounted_load_kwh` / `ev_total_planned_load_kwh` | **0** |
| `batteries_charged_kwh` | **0.631** |
| `estimated_battery_soc_pct` | 59.21 % (not full) |

The MILP planned **zero** EV energy for the slot (all three EV load fields are 0) and is charging the house battery (0.631 kWh) — yet `ev_charger_calculated_power` is 3815 W, so the Go-E is physically pulling ~3.6 kW for a charge the planner has already retracted.

## Root cause (two parts)

1. **Stale slot-start command is pinned (bug).** The per-slot EV power freeze (issue #738, `coordinator.py::_freeze_ev_charger_power_for_current_slot`) captures the charger command at slot start and restores it on every replan. When the planner flips from "charge EV" to "charge battery" mid-slot, the stale non-zero command survives because the freeze only guards magnitude, not the charge/no-charge decision.

2. **EV outranks the house battery for surplus PV (design).** In charge-past-target mode the EV is valued at `future_value_per_kwh = confidence_factor × mean(import_price, next 24h)` (issue #630), which can exceed the house battery's charge credit. The spec's priority table still states *"house battery fills first, then export, then EV gets remainder"*, but the #630 change lets the EV's speculative value outrank the battery's concrete scheduled need. The result is oscillation: the EV wins the surplus in one replan, the battery wins it in the next.

## Expected behaviour

- When the EV is at/above target SoC, surplus PV should fill the house battery first. The EV should only absorb surplus the battery cannot take.
[...]

Issue #615, opened 2026-06-27
feat: Session-aware EV demand — treat active EV charging as near-certain in optimiser


woopstar:
## Problem
The MILP treats EV load in all future slots with equal probability-based demand. When an EV is actively plugged in and drawing power, the next ~2 hours of EV demand should be treated as near-certain, not probabilistic.

## Proposal
When an EV is plugged in AND actively charging (in forced-draw modes: full power, solar+battery, or scheduled-and-active), the MILP should:
- Treat the next 2 hours of EV load as near-certain at the current charge power
- Beyond that window, fall back to the learned hourly probability model
- NOT grid-charge the house battery against this certain EV demand

## Implementation
- Read live EV state: `EVLiveState.is_charging`, `EVLiveState.power_w`
- Modify `planner/milp_optimizer.py` `solve_milp()`: when `ev_configs` are provided, check if any EV is actively charging; if so, set the first 2h of EV demand slots to the current charge power (instead of the probabilistic value from `EVConfig`)
- Add a `session_demand_kw: float | None` field to `EVConfig` (set by coordinator before calling MILP)
- Edit: `custom_components/hsem/models/ev_config.py` — add `session_demand_kw` field
- Edit: `custom_components/hsem/coordinator.py` — populate `session_demand_kw` from live EV power before building planner input

## Acceptance
- [ ] Active EV demand replaces probabilistic demand for first 2h
- [ ] Falls back to probabilistic model beyond 2h
- [ ] Does NOT grid-charge battery against certain EV demand
- [ ] Works for both primary and second EV
- [ ] `tox -e lint`, `tox -e typing` pass

PR #1013, opened 2026-09-15
fix(milp): pro-rate charge-past-target surplus bound for the live slot

### LINKED ISSUE #1012: fix(milp): charge-past-target surplus cap is not pro-rated for the live slot, so slot tails command the EV far above real PV surplus
## Description

A charge-past-target EV is supposed to charge from genuine PV surplus only. `docs/planner-spec.md:1210` states the guarantee explicitly:

> The constraint `ev_c[t]/charger_eff ≤ max(0, pv[t] − base_load[t])` ensures past-target EV charging **never** draws from the battery or grid — only genuine PV surplus that has nowhere else to go.

That guarantee is broken for the **live (partly elapsed) slot**, because the surplus bound is measured over the *full* slot while the resulting charger command is delivered over the *remaining* minutes.

**Root cause.** Both charge-past-target rows in `custom_components/hsem/planner/milp/_ev_constraints.py` use un-pro-rated full-slot surplus energy:

- surplus-only row — `_ev_constraints.py:248`: `surplus_kwh = max(pv_avail[t] - base_load[t], 0.0)`
- battery-first row (issue #775) — `_ev_constraints.py:269`: same expression

Every other link in the chain *is* pro-rated by `available_slot_hours[t]`:

- `_bounds.py:109` scales the `ev_c[t]` ceiling by `available_slot_hours[t] / slot_hours`
- `_ev_amp_lattice.py:304` ties `ev_c[t]` to `amps × available_slot_hours[t]`
- `_ev_power_writeout.py:88-96` converts the live slot's energy to a command by dividing by *remaining* hours

So the live slot's command can reach `full_slot_surplus / remaining_hours` instead of the correct `full_slot_surplus / slot_hours`.

**Impact.** Reproduced with a real user's configuration (86.5 kWh EV, 11 kW / 92 % / 1380 W minimum charger, 15-minute slots, house battery full, 600 W of genuine PV surplus):

```
genuine PV surplus = 600 W  (charger minimum = 1380 W -> should never start)

now=13:00    15 min left  command=     0W  import=0.000  non-PV draw=     0W
now=13:06     9 min left  command=     0W  import=0.000  non-PV draw=     0W
now=13:09     6 min left  command=  1380W  import=0.000  non-PV draw=   780W
now=13:12     3 min left  command=  2990W  import=0.000  non-PV draw=  2390W
now=13:14     1 min left  command=  8970W  import=0.000  non-PV draw=  8370W
```

Two compounding failures:

1. **The compression lifts the command over `charger_min_power_w`.** 600 W of surplus can never run a 1380 W charger. HSEM correctly refuses at the start of the slot, then starts it anyway in the tail — so a surplus that was never large enough to run the charger repeatedly does. With 15-minute slots this fires roughly four times an hour, every hour the EV sits above target.
2. **The published plan hides it.** `grid_import_kwh` stays `0.000` because the cost model believes a full slot's surplus is still available in that last minute. The import never appears in the plan or its cost.

Net effect: surplus is exported at the export price for most of the slot, then bought back at the import price (or drained from the house battery) during the tail burst.

Field report: an EV with target SoC 80 % charged 94 % → 100 % between 11:00 and 16:30 on a day when the house battery was full and PV surplus was modest. This is **not** the same defect as #988/#996 (unavailable SoC coerced to 0 %); it reproduces with a perfectly healthy SoC sensor.

## Possible Solution

Pro-rate the surplus bound for a partly elapsed slot in both charge-past-target rows, using the same `available_slot_hours[t] / slot_hours` scale that `_bounds.py`, `_session_window.py` and `_phase_fuse.py` already apply:

- Thread the existing `slot_hours` value from `_build_constraints` (`_constraints.py:215`) into `add_ev_and_session_constraint_rows`.
- Scale the RHS of the surplus-only row and the battery-first row by that fraction.

The fraction computation is currently inlined in three places (`_bounds.py:109`, `_session_window.py:109`, `_phase_fuse.py:143`). Per the repository's utility-centralisation rule it should become one canonical helper in `utils/units.py` rather than gaining a fourth copy, with the three existing call sites converted to it (pure move, no behaviour change).

Verified against the prototype: with the fix, a 600 W surplus never starts the 1380 W charger at any point in the slot, and a genuine 3 kW surplus still charges at ~2990 W regardless of how much of the slot has elapsed.

Out of scope (separate follow-up): the session grid-charge-prevention row (issue #615) at `_ev_constraints.py:317` compares full-slot `pv_avail[t]` against pro-rated session AC load — the same un-pro-rated shape, but it caps battery charging rather than EV charging and belongs to a different feature.

## Acceptance Criteria

- [ ] The surplus-only row and the battery-first row scale their RHS by the remaining fraction of a partly elapsed slot
- [ ] A charge-past-target EV whose genuine surplus is below `charger_min_power_w` is never commanded to charge, at any point within the slot
- [ ] A charge-past-target EV with genuine surplus above the minimum still charges at (and never above) the surplus power, independent of how much of the slot has elapsed
- [ ] A full-width future slot's behaviour is unchanged (scale is exactly 1.0)
- [ ] The remaining-fraction computation lives in one canonical helper, with the existing inline copies converted
- [ ] `docs/planner-spec.md` states that the surplus bound is pro-rated for the live slot, with an invariant covering it
- [ ] Tests added/updated covering the above
- [ ] `./scripts/quality.sh all` passes

## Suggested Prompt

```
Fix the charge-past-target surplus leak in the HSEM MILP.

Activate the hsem-pre-flight skill first, then the hsem-planner-change skill.

The bug: custom_components/hsem/planner/milp/_ev_constraints.py bounds
charge-past-target EV charging with max(pv_avail[t] - base_load[t], 0.0) — full-slot
forecast surplus energy — in both the surplus-only row (~line 248) and the
battery-first row (~line 269). But _bounds.py, _ev_amp_lattice.py and
_ev_power_writeout.py all pro-rate the live slot by available_slot_hours[t], and
_ev_power_writeout.py derives the charger command by dividing the slot's energy by
the REMAINING hours. So in the tail of a partly elapsed slot the command reaches
full_slot_surplus / remaining_hours instead of full_slot_surplus / slot_hours —
overshooting real surplus by up to 15x with 15-minute slots, lifting the command over
charger_min_power_w so the charger starts on surplus too small to run it, and
reporting grid_import_kwh = 0 while the site actually imports.

Fix: thread the existing slot_hours from _build_constraints (_constraints.py:215)
into add_ev_and_session_constraint_rows and scale both rows' RHS by
clamp(available_slot_hours[t] / slot_hours, 0, 1). The same fraction is inlined at
_bounds.py:109, _session_window.py:109 and _phase_fuse.py:143 — per the
utility-centralisation rule in CLAUDE.md, extract one canonical helper into
utils/units.py and convert those three call sites rather than adding a fourth copy.

Do NOT change the session grid-charge-prevention row (issue #615) at the bottom of
_ev_constraints.py — same shape, different feature, separate follow-up.

Add regression tests under tests/planner/ that solve the MILP with a charge-past-target
EV at several elapsed-time offsets within the live slot and assert the command never
exceeds the genuine surplus power, plus an unchanged-behaviour test for a full-width
slot. Use pytest.approx for all float assertions.

Update docs/planner-spec.md — the surplus-only section around line 1208 promises past-target
charging "never draws from the battery or grid"; document the pro-rating and add the
invariant. Then run ./scripts/quality.sh lint, typing, quality, test before opening the PR.
```

woopstar:
## Summary

Fixes #1012. A charge-past-target EV is supposed to charge from genuine PV surplus only — `docs/planner-spec.md` states it as a hard guarantee: past-target charging *"**never** draws from the battery or grid"*. That guarantee was broken for the **live (partly elapsed) slot**.

Both charge-past-target rows in `planner/milp/_ev_constraints.py` bounded `ev_c[t]` with `max(0, pv[t] − base_load[t])` — **full-width** forecast surplus energy. Every other link in the chain is pro-rated by `available_slot_hours[t]`:

- `_bounds.py:109` scales the `ev_c[t]` ceiling
- `_ev_amp_lattice.py:304` ties `ev_c[t]` to `amps × available_slot_hours[t]`
- `_ev_power_writeout.py:88-96` derives the charger command by dividing the live slot's energy by the hours that **remain**

So in a partly elapsed slot the command could reach `full_slot_surplus / remaining_hours` instead of `full_slot_surplus / slot_hours`.

## Impact

Reproduced with the reporter's configuration (86.5 kWh EV, 11 kW / 92 % / 1380 W minimum charger, 15-minute slots, house battery full, 600 W of genuine surplus):

| slot time left | command (before) | command (after) | genuine surplus |
| -------------- | ---------------- | --------------- | --------------- |
| 15 min         | 0 W              | 0 W             | 600 W           |
| 9 min          | 0 W              | 0 W             | 600 W           |
| 6 min          | **1380 W**       | 0 W             | 600 W           |
| 3 min          | **2990 W**       | 0 W             | 600 W           |
| 1 min          | **8970 W**       | 0 W             | 600 W           |

Two failures compound:

1. **The compression clears `charger_min_power_w`.** 600 W of surplus can never run a 1380 W charger. HSEM correctly refused at the start of the slot, then started it anyway in the tail — so a surplus that was never large enough to run the charger repeatedly did, roughly four times an hour with 15-minute slots.
2. **The published plan hid it.** `grid_import_kwh` stayed `0.000` because the cost model believed a full slot's surplus was still available in the final minute, so the import never appeared in the plan or its cost.

Field report that prompted this: an EV with target SoC 80 % charged 94 % → 100 % between 11:00 and 16:30 while the house battery was full and PV surplus was modest. This is **not** #988/#996 (unavailable SoC coerced to 0 %) — it reproduces with a perfectly healthy SoC sensor.

## Changes

1. **`planner/milp/_ev_constraints.py`** — both the surplus-only row and the battery-first row (issue #775) now bound against `surplus_remaining[t]`, the surplus the slot has still to deliver, via a shared `_remaining_surplus_kwh()` helper. `add_ev_and_session_constraint_rows()` takes `slot_hours`, threaded from `_build_constraints` (`_constraints.py:215`) where it already exists.
2. **`utils/units.py`** — new `remaining_slot_fraction(available_hours, slot_hours)`. The computation was already inlined in `_bounds.py:109`, `_session_window.py:109` and `_phase_fuse.py:143`; per the utility-centralisation rule in `CLAUDE.md` it becomes one canonical helper and those three call sites are converted rather than gaining a fourth copy. Pure move — no behaviour change at those sites.
3. **`docs/planner-spec.md`** — `surplus_remaining[t]` replaces the bare `max(0, pv[t] − base_load[t])` in all six places it appeared, the "Key constraint" section explains why the pro-rating is load-bearing, and four invariants are added.
4. **`.github/memories.md`** — the battery-first entry is corrected to `_ev_constraints.py` (it said `_constraints.py`, stale since the extraction) and gains the pro-rating rule; a new "Partly elapsed live slot" canonical-pattern entry points at `remaining_slot_fraction()`.

Every future slot scales by exactly `1.0`, so only the live slot changes behaviour.

## Regression tests

New `tests/planner/test_ev_past_target_surplus_prorating.py` (27 tests). Written against the unmodified bound first: **13 failed pre-fix**, **14 passed** (the helper unit tests and the don't-break guards — full-width slots, genuine surplus still charging). With the fix all 27 pass.

- Sub-minimum surplus never starts the charger, at 7 elapsed-time offsets through the slot
- The command never exceeds the genuine surplus power, at the same 7 offsets
- 3 kW of genuine surplus still runs the charger with 3 minutes left (guards against the fix becoming a blanket "never charge in the slot tail" rule)
- A full-width slot's command and energy are unchanged
- Planned past-target load never exceeds the remaining surplus energy, and `grid_import_kwh == 0` is therefore honest
- `remaining_slot_fraction()` parametrised over full/half/one-minute/exhausted/clamped-both-ways/degenerate-width inputs

## Verification

- `./scripts/quality.sh lint` ✅ / `typing` ✅ (0 mypy errors) / `quality` ✅ (0 pyright errors) / `test` ✅ — **3391 passed**
- `./scripts/quality.sh translations` ✅ — 0 errors (no user-facing strings changed)
- Planner-spec checklist: energy balance, SoC bounds, `winner.cost == final_output.cost`, slot identity, terminal-SoC accounting, no-action baseline, safety gates — all unaffected; the only semantic change is the tightened surplus bound, which the spec now documents.
- File-size limits: no touched file approaches 30 KB / 1000 lines.

## Out of scope

The session grid-charge-prevention row (issue #615) at the bottom of the same file compares full-slot `pv_avail[t]` against pro-rated session AC load — the same un-pro-rated shape. It caps *battery* charging rather than EV charging and belongs to a different feature, so it is deliberately untouched here; noted in #1012 for a follow-up.

## Acceptance criteria

- [x] The surplus-only row and the battery-first row scale their RHS by the remaining fraction of a partly elapsed slot
- [x] A charge-past-target EV whose genuine surplus is below `charger_min_power_w` is never commanded to charge, at any point within the slot
- [x] A charge-past-target EV with genuine surplus above the minimum still charges at (and never above) the surplus power, independent of elapsed time
- [x] A full-width future slot's behaviour is unchanged (scale is exactly 1.0)
- [x] The remaining-fraction computation lives in one canonical helper, with the existing inline copies converted
- [x] `docs/planner-spec.md` states that the surplus bound is pro-rated for the live slot, with invariants covering it
- [x] Tests added covering the above
- [x] `./scripts/quality.sh all` passes


woopstar:
**Correction — battery-side test reverted.**

Earlier today I added `test_battery_charge_never_exceeds_remaining_surplus` / `test_battery_still_absorbs_the_remaining_surplus` to this PR and described the pre-fix battery excess as drawn from the grid. That was wrong, and the test commit has been reverted.

With **no EV configured at all**, the same scenario plans 0.873 kWh of house-battery charge in the final minute of the live slot with `grid_import_kwh == 0`. The excess comes from the base model's full-width `pv[t]` bound, is independent of any EV, and is not a physical draw: the applier uses `batteries_charged_kwh` only as a zero/non-zero materiality signal, whereas the EV charger command is energy divided by remaining hours. The EV half of #1012 — what this PR fixes — is physical; the battery half is a planning artefact.

The assertion only passed because the battery-first row coupled the house battery to the EV's pro-rated budget, making battery behaviour depend on whether an EV is plugged in. #1015 removes that coupling.

This PR is back to the original, verified 27-test state; no production code changed.

Issue #1032, opened 2026-09-17
fix(planner): batteries_wait_mode slot retains MILP discharge after concentration


woopstar:
Reported by @AndersN76 in
https://github.com/woopstar/hsem/issues/978#issuecomment-5709334311 — filed as its
own issue with an independent reproduction against `main`.

## Description

`concentrate_discharge_on_expensive_slots()` relabels a non-retained discharge slot to
`batteries_wait_mode` but only clears the *charge* field
(`planner/discharge_scheduler.py:431-432`):

```python
s.recommendation = Recommendations.BatteriesWaitMode.value
s.batteries_charged_kwh = 0.0
```

It leaves `batteries_discharged_kwh`, `grid_import_kwh`, `grid_export_kwh`, the SoC
trajectory and the slot cost untouched.

For the **MILP candidate** that contradiction then survives the simulation.
`planner/candidate_selector.py` runs concentration at `:224` and *then* calls
`simulate_soc(...)` at `:235` with `milp_prepopulated=(candidate.name == CANDIDATE_MILP)`
(`:247`). In that mode `planner/soc_simulation.py:240-246` trusts the LP's energy values
verbatim:

```python
if milp_prepopulated:
    discharge = slot.batteries_discharged_kwh
```

The `batteries_wait_mode → discharge = 0.0` guard exists only at
`soc_simulation.py:270`, which sits inside the `elif net_demand > 0:` re-derivation
branch that prepopulated mode skips entirely.

**Scope:** non-MILP candidates self-heal, because they go through the re-derivation path
and hit the `:270` guard. Only the MILP candidate — the default planner path — publishes
the contradiction.

### Reported behaviour (v6.3.2, 15-minute slots, strict Wait)

During the active 06:30–06:45 slot the working-mode sensor carried:

| Field | Value |
| --- | --- |
| `recommendation` | `batteries_wait_mode` |
| `batteries_discharged_kwh` | 0.625 |
| `estimated_battery_capacity_kwh` | 2.125 |
| `estimated_battery_soc_pct` | 52.5 |
| `grid_import_kwh` | 0.216 |

Actual SoC was ~65%. Strict Wait was executed, so battery power went to 0 W and the
house imported from the grid while the plan still accounted for 0.625 kWh of discharge.
The following 06:45–07:00 slot repeated it with ~0.429 kWh. At 07:00 a fresh replan read
the real SoC and produced a consistent plan again — which is what makes the mode and
plan appear to jump between replans.

[...]

Issue #860, opened 2026-08-31
Triage: battery-schedule feature appears inert since MILP became the sole active planner


woopstar:
## Problem

The battery-schedule feature (`batteries_schedule_1/2/3` config, their
switches/time entities, `BatterySchedule`/`BatteryScheduleInput` models, and
the `apply_discharge_schedules` / `apply_charge_schedules` /
`apply_arbitrage_grid_charge` passes in `planner/engine_core.py:134-179`)
predates the MILP-based planner and appears to be functionally inert
whenever MILP is available (the default).

In MILP-only mode, `planner/candidate_generator.py` only produces `no_action`,
`passive`, and `milp` candidates (`candidate_generator.py:274-390`, the
schedule-consuming heuristic candidates are commented out at
`candidate_generator.py:288-323`). Both `no_action` and `passive` explicitly
clear all charge/discharge recommendations before scoring
(`planner/candidates/_mutations.py:39-80`), and MILP itself never reads
`battery_schedules` at all (confirmed via grep across `planner/milp/` and
`planner/milp_optimizer.py`). So whichever candidate wins, the
schedule-driven recommendations computed in `engine_core.py:134-179` never
reach the final selected plan.

This is already partially documented as intentional in
`docs/candidate-generation.md:33-48` and `docs/planner-spec.md:1783-1786`,
but the battery-schedule config options, switches, and time entities remain
fully exposed in the config flow and UI (`config_flow.py:604-624`,
`options_flow.py:403-423`, `flows/batteries_schedules.py`, `switch.py:61-71`)
with no indication to the user that they currently have no effect.

## Ask

Decide on one of:

1. **Remove** the battery-schedule feature entirely (config flow steps,
   models, switches/time entities, and the dead `apply_*_schedules` calls in
   `engine_core.py`) since MILP has fully superseded it.
2. **Re-wire** it as an opt-in fallback/diagnostic signal for when MILP is
   unavailable (e.g. scipy missing) — currently even the fallback candidates
   discard it, so this would require restoring it into `passive`/`no_action`
   or a new fallback candidate.
3. **Document only**: leave the code as-is but add a clear "currently has no
   effect while MILP is active" note in the UI/config flow and docs.

This issue is for triage/decision — implementation should happen in a
follow-up once the direction is chosen.

PR #1040, opened 2026-09-17
chore(planner): measure discharge concentration on the MILP candidate

### LINKED ISSUE #1036: chore(planner): determine whether discharge concentration should run on the MILP candidate
## Description

Follow-up question raised while fixing #1032, deliberately left out of scope there.

`concentrate_discharge_on_expensive_slots()` exists to thin discharge slots the battery
cannot serve. Its docstring states the reason plainly:

> `apply_optimization_strategy` marks *every* slot in a discharge window as
> `BatteriesDischargeWindowMode`, but the battery can only cover a fraction of them.

That is a description of the **seasonal fill**, not the MILP. The MILP allocates discharge
under exact SoC constraints and never over-allocates. Concentration's own estimate is
explicitly cruder (`discharge_scheduler.py:333-335`):

> The estimate within each day is conservative: it assumes the battery starts at full
> capacity and there is no incoming charge between discharge slots on the same day.

After #1032/#1033, concentration reserves LP-dispatched slots and only thins slots with no
solved discharge. On the MILP candidate, those remaining slots are exactly the ones
`apply_optimization_strategy` added where **the LP chose not to act**. So the open question
is whether concentration should run on the MILP candidate at all — and, separately, whether
the seasonal fill should be adding discharge-window labels to LP-idle slots in the first
place.

Measured from a production log (2026-09-16, 15-minute slots, 48h horizon): **327 clearings
across three candidates**, of which 3 were the LP-dispatched slots that caused #1032. The
remaining ~324 were seasonal-fill slots. That volume is worth understanding rather than
assuming.

There is a risk in the other direction too, noted in #1033: if concentration is currently
*load-bearing* for the MILP candidate — compensating for the LP over-allocating somewhere,
or for the seasonal fill adding too much — then removing it could surface a different
problem. That is why this is an investigation, not a pre-decided change.

## Possible Solution

Measure first, then decide. This should not start with a code change.

1. **Instrument.** #1033 added
   `[disch] concentrate: reserved N LP-dispatched slot(s)`. Add a companion count of how
   many *seasonal-fill* slots are cleared per candidate, so a log answers "what is
   concentration actually doing on the MILP candidate".
2. **Compare plan cost** with concentration enabled vs skipped for the MILP candidate
   only, across real cycles. The `hsem.export_diagnostics` dump carries a replayable
   `PlannerInput` (verified lossless — all 87 fields round-trip and reproduce a
   bit-identical plan), so this can be evaluated offline without touching a live system.
   See the backtest-harness issue for the replay shim.
3. **Then choose**, based on the numbers:
   - **Skip concentration for the MILP candidate** (`candidate.name == CANDIDATE_MILP`) if
     it never improves cost — the LP already solved the allocation.
   - **Stop the seasonal fill labelling LP-idle slots** on the MILP candidate, which would
     remove concentration's input rather than concentration itself. Note the ordering
     comment at `candidate_selector.py:204-213` explains why fill runs *before*
     concentration, so this path needs care.
   - **Keep as-is** if concentration measurably helps.

Whichever way it goes, the behaviour must be documented in `docs/planner-spec.md` — the
interaction between the LP, the seasonal fill and concentration on the MILP candidate is
currently only described in code comments, and one of those comments was wrong until
#1033 corrected it.

## Acceptance Criteria

- [ ] Log output shows, per candidate, how many LP-dispatched slots were reserved and how many seasonal-fill slots were cleared
- [ ] Plan cost is compared for the MILP candidate with concentration enabled vs skipped, over real replayed cycles rather than synthetic fixtures
- [ ] A decision is recorded in the issue with the supporting numbers before any behaviour change lands
- [ ] If behaviour changes, `docs/planner-spec.md` documents the LP / seasonal-fill / concentration interaction on the MILP candidate
- [ ] The #1032 invariant still holds — no `batteries_wait_mode` slot retains discharge
- [ ] Planner invariants still hold (energy balance, SoC bounds, cost identity)
- [ ] Tests added/updated covering the above
- [ ] `./scripts/quality.sh all` passes

## Suggested Prompt

```
Open question from #1032/#1033: should concentrate_discharge_on_expensive_slots() run on the
MILP candidate at all?

Its docstring says it exists because "apply_optimization_strategy marks *every* slot in a
discharge window as BatteriesDischargeWindowMode, but the battery can only cover a fraction
of them" — that describes the seasonal fill, not the MILP, which allocates under exact SoC
constraints and never over-allocates. Concentration's own per-day estimate is explicitly
cruder (discharge_scheduler.py:333-335: assumes the battery starts full with no incoming
charge). Since #1033 it reserves LP-dispatched slots, so on the MILP candidate it now only
thins slots the LP deliberately left idle.

Do NOT start by changing behaviour. Measure first:

1. Add a per-candidate count of seasonal-fill slots cleared, alongside the
   "[disch] concentrate: reserved N LP-dispatched slot(s)" line #1033 added.
2. Compare MILP-candidate plan cost with concentration enabled vs skipped across real
   replayed cycles. hsem.export_diagnostics dumps a lossless, replayable PlannerInput (all
   87 fields round-trip to a bit-identical plan), so this runs offline.
3. Record the decision with numbers in the issue before changing anything.

For reference, one production log showed 327 clearings across three candidates, only 3 of
which were LP-dispatched. Be aware of the opposite risk: concentration may be load-bearing,
compensating for the seasonal fill adding too much — in which case removing it surfaces a
different bug. Also read the ordering comment at candidate_selector.py:204-213, which
explains why the seasonal fill must run before concentration.

This is planner work: read docs/planner-spec.md first and run the hsem-pre-flight and
hsem-planner-change skills. Any behaviour change needs the spec updated plus invariant tests,
including that the #1032 invariant still holds. Then run ./scripts/quality.sh lint, typing,
quality and test. Do not merge without explicit permission.
```

woopstar:
## Summary

Answers the open question from #1032/#1033: **should `concentrate_discharge_on_expensive_slots()` run on the MILP candidate at all?**

**Measured answer: yes, keep it — but not for the reason its docstring gives.** No planner behaviour changes in this PR; it is instrumentation, a replay shim, tests and documentation.

Full numbers are recorded on the issue: https://github.com/woopstar/hsem/issues/1036#issuecomment-5712798189

**Branch:** `chore/1036-concentration-milp-measurement`

## Files changed

| File | Change |
| --- | --- |
| `custom_components/hsem/planner/discharge_scheduler.py` | Adds `ConcentrationStats`; `concentrate_discharge_on_expensive_slots` takes a logging-only `candidate_name` and returns the stats |
| `custom_components/hsem/planner/candidate_selector.py` | Passes `candidate_name=candidate.name` |
| `scripts/replay_planner_input.py` | **New** — rebuilds a `PlannerInput` from a diagnostics dump for offline replay |
| `tests/planner/test_concentration_on_milp.py` | **New** — 10 tests locking in the measured invariants |
| `docs/planner-spec.md` | New "On the MILP candidate" subsection under discharge concentration |

## What changed and why

**Instrumentation (issue step 1)** — the `DONE` log line now reads e.g. `candidate=milp kept=80 cleared=77 (seasonal_fill=77 other=0) lp_reserved=19`, so the split #1033 made relevant is visible per candidate without re-deriving it. The per-slot clearing line also records the label it cleared (`was=…`).

**Replay shim (issue step 2)** — `scripts/replay_planner_input.py` is the inverse of `_planner_input_to_dict`, so a real `hsem.export_diagnostics` dump can be replayed offline. It returns a `ReplayReport` naming every field it had to drop or default; a replay that silently discarded inputs would produce numbers that look real and are not.

**Documentation** — the spec now records what concentration actually does on the MILP candidate, with its own test invariants.

## Findings

**Production scale** (159 log files, 1.4 GB, 2026-08-31 → 09-13 — 15,183 passes over 5,061 cycles): 62.0 % of slots entering concentration are cleared, 99.1 % of passes clear at least one slot, median 237 clearings per cycle. Consistent with the 327 quoted in the issue. These logs predate #1033 (no reservation lines), so they establish volume, not the post-fix split.

**The decisive A/B** — real production input, house load scaled so the per-day budget is genuinely exceeded:

| load × | MILP cleared | of which seasonal-fill | Δ plan cost |
| --- | --- | --- | --- |
| 2.0 | 2 | 2 | 0.0000 |
| 3.0 | 41 | 41 | 0.0000 |
| 5.0 | 77 | 77 | 0.0000 |
| 8.0 | 93 | 93 | 0.0000 |
| 12.0 | 98 | 98 | 0.0000 |

- 100 % of slots cleared on the MILP candidate are seasonal-fill window slots — post-#1032 it never touches an LP-dispatched slot.
- Plan cost is **bit-identical**, not within-epsilon: `total_cost` and `score` compare exactly equal and **zero** slots differ in any energy field. Expected — the MILP candidate is simulated with `milp_prepopulated=True`, so energy is trusted verbatim and labels carry no price.
- The **only** difference is the published label: up to 98 slots flipping `batteries_wait_mode` ↔ `batteries_discharge_window_mode`.

**Why that makes it load-bearing.** `batteries_discharge_window_mode` executes as `MaximizeSelfConsumption` (`applier.py:430-434`) — the firmware may discharge to cover house load. `batteries_wait_mode` holds at 0 W unless the opt-in `SelfConsumptionWithReserve` is configured (`applier.py:448+`). Skipping concentration on the MILP candidate would let the firmware drain the battery in ~77 slots the LP **deliberately left idle**. Plan cost cannot see this — which is exactly why a cost-only comparison would have answered "remove it" and been wrong.

So concentration's real job on the MILP candidate is:

> suppress the seasonal fill's discharge-window labels on LP-idle slots, so those slots execute as a hold rather than as self-consumption.

## Tests added

`tests/planner/test_concentration_on_milp.py` — 10 tests:

- Instrumentation: stats reported per candidate, `cleared == seasonal_fill + other`, early return still reports.
- Measured properties: only seasonal-fill slots cleared on the MILP candidate (parametrized over load ×3/×5/×8); skipping leaves cost and every energy field identical; skipping flips labels wait → window.
- Invariants: the #1032 rule holds with and without concentration; SoC stays in bounds either way.

## Test and lint results

```
./scripts/quality.sh lint     → 0
./scripts/quality.sh typing   → 0
./scripts/quality.sh quality  → 0
./scripts/quality.sh test     → 3571 passed
```

## Known limitations and open questions

- I had **one** real diagnostics dump. On that cycle concentration clears nothing on any candidate (per-day need 2.21/4.58 kWh vs a 9.5 kWh budget), so it yields a zero cost delta that is **not** evidence — it never acted. The table above therefore scales house load on the real input: real prices, real PV, real battery config, synthetic load multiplier. Labelled as semi-synthetic rather than presented as replayed cycles.
- The dump is v6.3.2 replayed against current `main`, so it is not bit-identical to what v6.3.2 produced: `battery_schedules` and `extra` are dropped (the former removed in #860) and `live_solar_production_available` defaults. The shim reports all three rather than hiding them.
- The production logs predate #1033, so their 62 % clear rate includes the LP-dispatched clearings that #1032 fixed.
- Follow-up filed as **#1041**: the seasonal fill arguably should not be labelling LP-idle slots at all. Measured asymmetry — winter assigns `BatteriesWaitMode` directly and needs 0 clearings; summer assigns the window label and needs 5–12. Deliberately not attempted here, since it touches the fill-before-concentration ordering.

## Required configuration changes

None. No config flow, entity, service or user-facing string changed, so no translation updates are needed.

## Acceptance criteria

- [x] Log shows per candidate how many LP-dispatched slots were reserved and how many seasonal-fill slots were cleared
- [x] Plan cost compared for the MILP candidate with concentration enabled vs skipped
- [x] Decision recorded in the issue with supporting numbers before any behaviour change
- [x] `docs/planner-spec.md` documents the LP / seasonal-fill / concentration interaction on the MILP candidate
- [x] #1032 invariant still holds — asserted both with and without concentration
- [x] Planner invariants still hold (energy fields unchanged, cost identity, SoC bounds)
- [x] Tests added (10 new)
- [x] `./scripts/quality.sh all` passes
- [ ] No behaviour change — deliberately, per the issue

## Note for the reviewer

`discharge_scheduler.py` is now 28,825 bytes, under the 30 KB limit but close.

The real production logs and diagnostics dump used for the measurement live under `logs/`, which is gitignored — they contain real user data and are not part of this commit.

Fixes #1036


woopstar:
**Note for merge order:** #1042 (issue #1041) is stacked on this branch — it builds on the `candidate_name=`/`ConcentrationStats` wiring here and rewrites the "On the MILP candidate" spec subsection this PR adds.

Two consequences:
- **Retarget #1042 to `main` before merging this PR**, otherwise merging with `--delete-branch` will auto-close it.
- The `test_skipping_changes_labels_from_wait_to_window` test added here is intentionally replaced in #1042: once the fill stops producing discharge-window labels on LP-idle slots, skipping concentration on the MILP candidate changes nothing. The finding in this PR still stands — it is *why* #1041 was worth doing.

